PR #2, opened 2020-11-02
Initial PR for Review and Comments

devngc:
Hi @mostaphaRoudsari & @chriswmackey, 
in togeometry, you'll see a code block is repeated. I kept it that way for better readability from my point of view.


honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
+import rhino3dm
```
mostaphaRoudsari, line 1: We need a couple of lines here to explain what this module is supposed to do.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+import uuid
+from togeometry import to_face3d
```
mostaphaRoudsari, line 3: for relative import you should us . before the module name. Depending on the operating system this can fail.

```suggestion
from .togeometry import to_face3d
```

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+from tojson import publish_json
```
mostaphaRoudsari, line 4:
```suggestion
from .tojson import publish_json
```

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+try:  # import the honeybee-radiance extension
+    from honeybee_radiance.lib.modifiers import modifier_by_identifier
+except ImportError as e:
+    if len(rad_mod_) != 0:
+        raise ValueError('rad_mod_ has been specified but honeybee-radiance '
+                         'has failed to import.\n{}'.format(e))
+
+
+# The path here is the path to the rhinofile in .3dm format
+path = "D:\Github\Work-in-Progress\RhinoToHBJSON\sample.3dm"
```
mostaphaRoudsari, line 41: Should we change the content of this module to a function and call the function?

You can't have a path hard-coded like this.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+file = rhino3dm.File3dm.Read(path)
```
mostaphaRoudsari, line 42: This will overwrite python's `file` object. Use a different name for the variable.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+hb_faces = []
+
+# Gathering layer information from the rhino file
+layers = file.Layers
+# Layer names in the Rhino file
+layer_names = [layer_name.Name for layer_name in layers]
+# Layer Indexes for the layers in the Rhino file
+layer_indexes = [layer_name.Index for layer_name in layers]
+# A dictionary with the layer names as keys and layer indexes as values
+layer_dict = dict(zip(layer_names, layer_indexes))
```
mostaphaRoudsari, line 55: You can do all of this in one go:

```suggestion
layer_dict = {layer_name.Name: layer_name.Index for layer_name in file.Layers}
```

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+
+    Args:
+        layer (A string): Name of the Rhino layer
+
+    Returns:
+        hb_type: A Honeybee face type object
+        hb_face_module: A Honeybee class. It is either of Face, Shade, Aperture, or Door
+    """
+    hb_type = " "  # Variable holding a Honeybee face type
+    hb_face_module = " "  # Variable holding a Honeybee module name
```
mostaphaRoudsari, line 70: Why do you need this? It will only happen in the final `else` check. I would just use it there.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+layers = file.Layers
+# Layer names in the Rhino file
+layer_names = [layer_name.Name for layer_name in layers]
+# Layer Indexes for the layers in the Rhino file
+layer_indexes = [layer_name.Index for layer_name in layers]
+# A dictionary with the layer names as keys and layer indexes as values
+layer_dict = dict(zip(layer_names, layer_indexes))
+
+
+def layer_name_to_hb_objects(layer):
```
mostaphaRoudsari, line 58: I think we can re-write this method to take advantage of a dictionary object for mapping the layers and also let the user overwrite the default layer structure with passing a dictionary as options.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+        hb_type = None
+        hb_face_module = Shade
+    if layer == "aperture":
+        hb_type = None
+        hb_face_module = Aperture
+    if layer == "door":
+        hb_type = None
+        hb_face_module = Door
+    if layer == "Default":
+        pass
```
mostaphaRoudsari, line 94: As I said this is probably better to be implemented as a mapper/dictionary but in case you want to keep it as is, all the checks except for the first one should be `elif`.

Also for strings we us `'` instead of `"`.

```suggestion
    if layer == 'wall':
        hb_type = face_types.wall
        hb_face_module = Face
    elif layer == 'floor':
        hb_type = face_types.floor
        hb_face_module = Face
    elif layer == 'airwall':
        hb_type = face_types.air_boundary
        hb_face_module = Face
    elif layer == 'shade':
        hb_type = None
        hb_face_module = Shade
    elif layer == 'aperture':
        hb_type = None
        hb_face_module = Aperture
    elif layer == 'door':
        hb_type = None
        hb_face_module = Door
    elif layer == 'Default':
        pass
```

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+
+    return (hb_type, hb_face_module)
+
+
+# Look at Rhino layers and
+for layer in layer_names:
+    hb_face_type = layer_name_to_hb_objects(layer)[0]
+    hb_face_module = layer_name_to_hb_objects(layer)[1]
```
mostaphaRoudsari, line 102:
```suggestion
    hb_face_type, hb_face_module = layer_name_to_hb_objects(layer)
```

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+    layer_name = layer
```
mostaphaRoudsari, line 103: Why do you need this? You can just call the `layer` variable `layer_name` when you start the for loop.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # Gathering planar geometries from the Rhino layer
+    rhino_faces = [object for object in file.Objects if object.Attributes.LayerIndex ==
+                   layer_dict[layer_name]]
```
mostaphaRoudsari, line 107: Having two loops like this might end up being pretty heavy for larger Rhino files. At the same time it might save us some memory. Depending on how slow it gets we might want to load all the objects in one-go and map them to a dictionary.

Also are you sure the 3dm library doesn't have a `select by layer` method that you can use instead of filtering it yourself?

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # Creating face names
+    face_names = []
+    for count, geo in enumerate(rhino_faces):
+        # If there's a user given name, use it
+        # TODO: Quality check for user input name will have to be created here
+        if len(geo.Attributes.Name) > 0:
```
mostaphaRoudsari, line 114: Assuming Name is a string then you can just do:

```suggestion
        if geo.Attributes.Name:
```

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+    hb_face_module = layer_name_to_hb_objects(layer)[1]
+    layer_name = layer
+
+    # Gathering planar geometries from the Rhino layer
+    rhino_faces = [object for object in file.Objects if object.Attributes.LayerIndex ==
+                   layer_dict[layer_name]]
+
+    # Creating face names
+    face_names = []
+    for count, geo in enumerate(rhino_faces):
```
mostaphaRoudsari, line 111: Am I right that you never use `count`?

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+        # If there's a user given name, use it
+        # TODO: Quality check for user input name will have to be created here
+        if len(geo.Attributes.Name) > 0:
+            face_names.append(geo.Attributes.Name)
+        # Use auto generated name
+        else:
+            name = layer_name + str(uuid.uuid4())[:8]
+            face_names.append(name)
```
mostaphaRoudsari, line 119: I personally think his is one of those cases that a single line code is cleaner but what you have here should also work.

```Python
face_names = [geo.Attributes.Name for geo in rhino_faces if geo.Attributes.Name else layer_name + str(uuid.uuid4())[:8]]
```

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # for each rhino geometry gathered
+    for i in range(len(rhino_faces)):
+        face = rhino_faces[i].Geometry
```
mostaphaRoudsari, line 123: This is how you would do a loop in C# or Javascript but not in Python.

```suggestion
    for face_obj in rhino_faces:
        face = face_obj .Geometry
```

I just comment on this one but you get the concept.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+                    '{}_{}'.format(face_names[i], i)), lb_face[j])
+            if hb_face_module == Aperture:
+                hb_face = hb_face_module(clean_and_id_string(
+                    '{}_{}'.format(face_names[i], i)), lb_face[j])
+            if hb_face_module == Door:
+                hb_face = hb_face_module(clean_and_id_string(
+                    '{}_{}'.format(face_names[i], i)), lb_face[j])
+            # Assigning a name to the Honeybee Face
+            hb_face.display_name = '{}_{}'.format(face_names[i], i)
+            hb_faces.append(hb_face)
```
mostaphaRoudsari, line 142: This `i` and `j` studd really needs to be replaced by how it is done in Python. That will make the code much more readable.

I also suggest to separate this part in a separate re-usable function.

Finally, if we are not concerned about Python 2 you may want to start using f-strings which will make the code easier to read.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+
+# Write the Honeybee faces in a JSON file
+publish_json(hb_faces)
```
mostaphaRoudsari, line 145: This will probably change when we refactor this module to be a number of re-usable functions.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,96 @@
...
-    pass
+    Args:
+        geo (A Rhino Brep, Surface or Mesh): This will be converted into a list
+        of Ladybug Face3D.
+        meshing_parameters (string): Optional Rhino Meshing Parameters
+        to describe how curved faces should be converted into planar elements.
+        If None, Rhino3dm's Default Meshing Parameters will be used.
+    """
 
+    # This is a container
```
mostaphaRoudsari, line 36: I don't think we need this comment.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,96 @@
...
 
-def to_linesegment3d(line):
-    """Ladybug LineSegment3D from Rhino LineCurve."""
-    pass
+    Args:
+        geo (A Rhino Brep, Surface or Mesh): This will be converted into a list
+        of Ladybug Face3D.
+        meshing_parameters (string): Optional Rhino Meshing Parameters
+        to describe how curved faces should be converted into planar elements.
+        If None, Rhino3dm's Default Meshing Parameters will be used.
```
mostaphaRoudsari, line 33:
```suggestion
        geo (A Rhino Brep, Surface or Mesh): This will be converted into a list
            of Ladybug Face3D.
        meshing_parameters (string): Optional Rhino Meshing Parameters
            to describe how curved faces should be converted into planar elements.
            If None, Rhino3dm's Default Meshing Parameters will be used.
```

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,96 @@
...
+    return Point3D(point.X, point.Y, point.Z)
+
+
+def to_vector3d(vector):
+    """Ladybug Vector3D from Rhino Vector3d."""
+    return Vector3D(vector.X, vector.Y, vector.Z)
+
 
+def to_face3d(geo, meshing_parameters=None):
+    """This method converts a rhino face into a Ladybug Face3D object. The
```
mostaphaRoudsari, line 25: This docstring is different from what the function really does.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,96 @@
...
+            if len(face) == 4:
+                all_verts = (pts[face[0]], pts[face[1]],
+                             pts[face[2]], pts[face[3]])
+            else:
+                all_verts = (pts[face[0]], pts[face[1]], pts[face[2]])
+            # Create Ladybug Face3D objects based on tuples of vertices
+            faces.append(Face3D(all_verts))
 
+    # If it's an Extrusion
+    if isinstance(geo, rhino3dm.Extrusion):
```
mostaphaRoudsari, line 55:
```suggestion
    elif isinstance(geo, rhino3dm.Extrusion):
```

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,96 @@
...
+        # Convert it into a Mesh first
+        mesh = geo.GetMesh(rhino3dm.MeshType.Default)
+        if isinstance(mesh, rhino3dm.Mesh):
```
mostaphaRoudsari, line 58: You don't need this check. Doesn't line 57 always return a mesh?

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,96 @@
...
+            # Get all the vertices
+            pts = [to_point3d(mesh.Vertices[i])
+                   for i in range(len(mesh.Vertices))]
```
mostaphaRoudsari, line 61: Can't we just take the vertices here:

```suggestion
            pts = [to_point3d(v) for v in mesh.Vertices]
```

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,96 @@
...
+            for j in range(len(mesh.Faces)):
+                face = mesh.Faces[j]
```
mostaphaRoudsari, line 63: Same here. just use the object whenever possible.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,96 @@
...
+                if len(face) == 4:
+                    all_verts = (pts[face[0]], pts[face[1]],
+                                 pts[face[2]], pts[face[3]])
+                else:
+                    all_verts = (pts[face[0]], pts[face[1]], pts[face[2]])
+                # Create Ladybug Face3D objects based on tuples of vertices
+                faces.append(Face3D(all_verts))
```
mostaphaRoudsari, line 70: You should separate and re-use this part of the code. Basically you can do a check at the start and if it is not a mesh then generate the mesh first and then pass the mesh to this re-usable method that creates a LBMesh from the Rhino mesh.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,96 @@
...
 
-def to_polyface3d(geo):
-    """A Ladybug Polyface3D object from a Rhino Brep."""
-    pass
+    # If it's a Brep
+    if isinstance(geo, rhino3dm.Brep):
```
mostaphaRoudsari, line 73:
```suggestion
    elif isinstance(geo, rhino3dm.Brep):
```

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,96 @@
...
+        # Convert it into a list of Meshes
+        meshes = [geo.Faces[f].GetMesh(rhino3dm.MeshType.Any) for f in range(
```
mostaphaRoudsari, line 75: same comments as above.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,96 @@
...
+                       for i in range(len(mesh.Vertices))]
+                for j in range(len(mesh.Faces)):
+                    face = mesh.Faces[j]
+                    if len(face) == 4:
+                        all_verts = (pts[face[0]], pts[face[1]],
+                                     pts[face[2]], pts[face[3]])
+                    else:
+                        all_verts = (pts[face[0]], pts[face[1]], pts[face[2]])
+                    # Create Ladybug Face3D objects based on tuples of vertices
+                    faces.append(Face3D(all_verts))
```
mostaphaRoudsari, line 95: You need a final check here for unsupported objects which are not Mesh, Extrusion or Brep to give an error or a warning to use so they know those object types are not supported.

honeybee_3dm/tojson.py
```diff
@@ -0,0 +1,39 @@
...
+    from honeybee.model import Model
+    from honeybee.config import folders
+except ImportError as e:
+    raise ImportError('\nFailed to import honeybee:\n\t{}'.format(e))
+
+import os
+import json
+
+
+def publish_json(_hb_objs, _name_=None, _folder_=None, indent_=None, abridged_=False, _dump=True):
```
mostaphaRoudsari, line 11: I don't think that we use the term `publish` in our code-base at all. It should probably change to `to_json`. @chriswmackey is the right person to confirm. If he is fine with publish then ignore this comment.

honeybee_3dm/tojson.py
```diff
@@ -0,0 +1,39 @@
...
+
```
mostaphaRoudsari, line 12: Missing docstring.

honeybee_3dm/tojson.py
```diff
@@ -0,0 +1,39 @@
...
+    # set the component defaults
```
mostaphaRoudsari, line 13: component?

honeybee_3dm/tojson.py
mostaphaRoudsari, line 11: This is how we name arguments inside the Grasshopper components to clarify which ones are required. In the python code we don't use `_` before and after the argument names.

honeybee_3dm/tojson.py
```diff
@@ -0,0 +1,39 @@
...
+        obj_dict = {}
+        for obj in _hb_objs:
+            try:
+                obj_dict[obj.identifier] = obj.to_dict(abridged=abridged)
+            except TypeError:  # no abridged option
+                obj_dict[obj.identifier] = obj.to_dict()
+
+    # write the dictionary into a file
+    with open(hb_file, 'w') as fp:
+        json.dump(obj_dict, fp, indent=indent)
```
mostaphaRoudsari, line 39: I _feel_ for this library we probably want to have ways to create a `Model` which we can then convert to an energy model or a Radiance folder. Not sure how useful writing the separate faces to a JSON file will be for what we need at this point.

mostaphaRoudsari (commented):
Hi @devngc, I left my comments. It might look too many for the amount of code but this is expected for the first couple of PRs. I suspect this will change after the couple of first PR. Thank you for putting this PR together. 😀

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+    from honeybee.face import Face
+    from honeybee.shade import Shade
+    from honeybee.aperture import Aperture
+    from honeybee.door import Door
+    from honeybee.facetype import face_types
+    from honeybee.boundarycondition import boundary_conditions
+    from honeybee.typing import clean_and_id_string
+    from honeybee.facetype import face_types
+except ImportError as e:
+    raise ImportError('\nFailed to import honeybee:\n\t{}'.format(e))
```
chriswmackey, line 16: You don't need to put this all in a try/except. People will still get a pretty helpful error message without it. And the only reason why the `ladybug_rhino` package has this is because we are running it in IronPython where dependencies are a bot more tenuous than in cPython.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+
+try:
+    from ladybug_geometry.geometry3d.face import Face3D
+
+except ImportError as e:
+    raise ImportError('\nFailed to import ladybug_geometry:\n\t{}'.format(e))
```
chriswmackey, line 22: You also don't need the try/except here.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+try:  # import the honeybee-energy extension
+    from honeybee_energy.lib.constructions import opaque_construction_by_identifier
+except ImportError as e:
+    if len(ep_constr_) != 0:
+        raise ValueError('ep_constr_ has been specified but honeybee-energy '
+                         'has failed to import.\n{}'.format(e))
```
chriswmackey, line 30: Let's get rid of all these try/excepts and change the dependency in the requirements.txt to `lbt-honeybee` instead of `honeybee-core`

honeybee_3dm/face.py
chriswmackey: This should definitely be an input argument to a function

honeybee_3dm/face.py
```diff
@@ -0,0 +1,145 @@
...
+        hb_type: A Honeybee face type object
+        hb_face_module: A Honeybee class. It is either of Face, Shade, Aperture, or Door
+    """
+    hb_type = " "  # Variable holding a Honeybee face type
+    hb_face_module = " "  # Variable holding a Honeybee module name
+
+    if layer == "roof":
+        hb_type = face_types.roof_ceiling
+        hb_face_module = Face
+    if layer == "wall":
```
chriswmackey, line 75: `elif` statements are probably better here as they will prevent the running of extra unnecessary conditional statements.

chriswmackey (commented):
I added a few more comments. We can discuss more on the call.

devngc:
Thank you @mostaphaRoudsari and @chriswmackey  for taking the time to post comments. I will comply with the comments. I can also share that I removed the repeated code blocks from togeometry. The next PR will have fewer lines of code to review.

honeybee_3dm/face.py
devngc: Complied

honeybee_3dm/face.py
devngc: To be discussed on the call. I do not understand how requirements.txt is being used and updated.

honeybee_3dm/face.py
devngc: Complied.

honeybee_3dm/face.py
devngc: Complied.

honeybee_3dm/tojson.py
devngc: To be discussed on the call.

honeybee_3dm/tojson.py
devngc: Complied.

honeybee_3dm/tojson.py
devngc: Complied.

honeybee_3dm/tojson.py
devngc: Docstring provided.

honeybee_3dm/tojson.py
devngc: Revised.

honeybee_3dm/face.py
devngc: Revised.

honeybee_3dm/face.py
devngc: Complied.

honeybee_3dm/face.py
devngc: Complied.

honeybee_3dm/face.py
devngc: Complied.

honeybee_3dm/face.py
devngc: Complied.

devngc:
This draft PR is not ready for review yet

honeybee_3dm/face.py
devngc: @mostaphaRoudsari, I did not want to use these studs but I observed that there's a difference in list length when I traverse a list pythonic way vs using these studs.

mostaphaRoudsari: Interesting! That should not happen. We can check the code together.

honeybee_3dm/face.py
devngc: These are variables created within the scope of the function. If I don't declare them, it doesn't work. The other way is to declare them as `global` variables, which I don't think we should do.

honeybee_3dm/togeometry.py
devngc: As mentioned above. This does not work. An iterator is needed. I will demo this on the call.

honeybee_3dm/togeometry.py
devngc: As mentioned above. This does not work. An iterator is needed. I will demo this on the call.

honeybee_3dm/togeometry.py
devngc: Yes. This is the repeated code that I alluded to. I have removed duplicate copies now and created a function for it.

honeybee_3dm/face.py
devngc: Yes. I did not find a `select by layer` method

honeybee_3dm/face.py
devngc: This does not work. An iterator is needed. I will demo this on the call.

honeybee_3dm/face.py
devngc: Revised

honeybee_3dm/togeometry.py
devngc: This function is removed altogether.

honeybee_3dm/togeometry.py
devngc: As mentioned above. This does not work. An iterator is needed. I will demo this on the call.

devngc:
Hi @mostaphaRoudsari,
This PR is ready for review.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,112 @@
...
+                    lb_face = brep_to_face3d(rh_face)
+
+                elif rh_face.ObjectType == rhino3dm.ObjectType.Extrusion:
+                    lb_face = extrusion_to_face3d(rh_face)
+
+                elif rh_face.ObjectType == rhino3dm.ObjectType.Mesh:
+                    lb_face = mesh_to_face3d(rh_face)
+                else:
+                    print(
+                        f'There are objects on layer "{layer}" that this library does not support.')
```
mostaphaRoudsari, line 82: You should be more specific here. And I would suggest to use Python's warning module instead of printing.

```suggestion
                    warnings.warn(f'{rh_face.ObjectType} object type is not supported.')
```

honeybee_3dm/face.py
```diff
@@ -0,0 +1,112 @@
...
+
+                    elif hb_face_module == Aperture:
+                        hb_face = hb_face_module(clean_and_id_string(
+                            '{}_{}'.format(face_names[count], count)), face_obj)
+
+                    elif hb_face_module == Door:
+                        hb_face = hb_face_module(clean_and_id_string(
+                            '{}_{}'.format(face_names[count], count)), face_obj)
+                    else:
+                        pass
```
mostaphaRoudsari, line 103: I assume since you have already checked the types up there we should never get here and you can remove this.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,112 @@
...
+                        hb_face = hb_face_module(clean_and_id_string(
+                            '{}_{}'.format(face_names[count], count)), face_obj)
+
+                    elif hb_face_module == Aperture:
+                        hb_face = hb_face_module(clean_and_id_string(
+                            '{}_{}'.format(face_names[count], count)), face_obj)
+
+                    elif hb_face_module == Door:
+                        hb_face = hb_face_module(clean_and_id_string(
+                            '{}_{}'.format(face_names[count], count)), face_obj)
```
mostaphaRoudsari, line 101: What is the difference between these 3 checks. If we name them all the same why are you separating one from another?

honeybee_3dm/room.py
```diff
@@ -0,0 +1,86 @@
...
+from honeybee.facetype import get_type_from_normal
+from honeybee.typing import clean_and_id_string
+from ladybug_geometry.geometry3d.face import Face3D
+from ladybug_geometry.geometry3d.polyface import Polyface3D
+
+
+def to_room(path):
+    """This function looks up a rhino3dm file, converts the objects
+    on the layer name "room" to Honeybee Room objects, and writes them
+    to a json file.
```
mostaphaRoudsari, line 26: Same comments as the other module.

honeybee_3dm/room.py
```diff
@@ -0,0 +1,86 @@
...
+    # TODO create a quality check for the file path
+
+    rhino3dm_file = rhino3dm.File3dm.Read(path)
+    tolerance = rhino3dm_file.Settings.ModelAbsoluteTolerance
+
+    layer_dict = {
+        layer.Name: layer.Index for layer in rhino3dm_file.Layers}
+
+    volumes = [object for object in rhino3dm_file.Objects if object.Attributes.LayerIndex ==
+               layer_dict["room"]]
```
mostaphaRoudsari, line 40: same comments about object.

honeybee_3dm/room.py
```diff
@@ -0,0 +1,86 @@
...
+            check.append(True)
+        elif geo.ObjectType == rhino3dm.ObjectType.Brep and geo.IsSolid == True:
+            check.append(True)
+        elif geo.ObjectType == rhino3dm.ObjectType.Mesh and geo.IsClosed == True:
+            check.append(True)
+
+    assert (len(check) == len(
+        volumes)), 'On the "rooms" layer, you must only have either closed brep,' \
+        'or closed extrusions, closed meshes.' \
+        ' Please make this change in the rhino file and try again'
```
mostaphaRoudsari, line 55:
```suggestion
    for geo in volumes:
        geo = geo.Geometry
        if geo.ObjectType == rhino3dm.ObjectType.Extrusion and geo.IsSolid == True:
            continue
        elif geo.ObjectType == rhino3dm.ObjectType.Brep and geo.IsSolid == True:
            continue
        elif geo.ObjectType == rhino3dm.ObjectType.Mesh and geo.IsClosed == True:
            continue
        else:
            raise ValueError(
                'On the "rooms" layer, you must only have either closed brep, '
                'or closed extrusions, closed meshes. '
                'Please make this change in the rhino file and try again'
           )
```

honeybee_3dm/room.py
```diff
@@ -0,0 +1,86 @@
...
+            lb_faces = brep_to_face3d(rh_solid)
+
+        elif rh_solid.ObjectType == rhino3dm.ObjectType.Extrusion:
+            lb_faces = extrusion_to_face3d(rh_solid)
+
+        elif rh_solid.ObjectType == rhino3dm.ObjectType.Mesh:
+            lb_faces = mesh_to_face3d(rh_solid)
+        else:
+            print(
+                f'There are objects on the layer named "room" that this library does not support.')
```
mostaphaRoudsari, line 77: Same comments as the one in to_face module.

I think you probably want to add a `to_face3d` function that takes all the object types and returns the lb_faces. That way you can re-use the code between these modules.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,93 @@
...
+        face = mesh.Faces[j]
+        if len(face) == 4:
+            all_verts = (pts[face[0]], pts[face[1]],
+                         pts[face[2]], pts[face[3]])
+        else:
+            all_verts = (pts[face[0]], pts[face[1]], pts[face[2]])
+
+        faces.append(Face3D(all_verts))
+
+    return faces
```
mostaphaRoudsari, line 60: Should we return a Ladybug Mesh instead of a list of faces?

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,93 @@
...
-    pass
+    Returns:
+        A list : A list of Ladybug Face3D objects
+    """
+    faces = []
+    for i in range(len(brep.Faces)):
+        mesh = brep.Faces[i].GetMesh(rhino3dm.MeshType.Any)
+        faces.extend(mesh_to_face3d(mesh))
 
+    return faces
```
mostaphaRoudsari, line 77: Should we return a PolyFace here?

honeybee_3dm/tojson.py
mostaphaRoudsari: I still think we don't need this module. User can use the methods provided in honeybee library to convert the geometries to json or hbjson as needed.

tests/face_test.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+import rhino3dm
+from pathlib import Path
+from honeybee.face import Face
+from honeybee_3dm.face import to_face
+
+
+def test_room():
+    path = str((Path.cwd()).joinpath('tests', 'test.3dm'))
```
mostaphaRoudsari, line 9: since we always run the tests from the root folder you can use relative path and keep it simple:

https://github.com/ladybug-tools/honeybee-energy/blob/95a436c0f6757b7ab8be72655d801fc936fca47e/tests/simulation_sizing_test.py#L25

I also suggest to put the 3dm file in a subfolder like `assets`.

tests/face_test.py
```diff
@@ -0,0 +1,11 @@
...
+    room = to_face(path)[0]
+    assert isinstance(room, Face)
```
mostaphaRoudsari, line 11: You probably want more checks to make sure it has been translated correctly. Right now you are just checking that it has been translated.

tests/module_test.py
```diff
@@ -0,0 +1,11 @@
...
+
+from honeybee_3dm.room import to_room
+from honeybee_3dm.face import to_face
+from honeybee_3dm.togeometry import mesh_to_face3d
+
+
+def test_import():
+    assert to_room
+    assert to_face
+    assert mesh_to_face3d
```
mostaphaRoudsari, line 11: unless we have some better tests to check, I think we can remove this file. These tests are not really useful.

tests/room_test.py
```diff
@@ -0,0 +1,11 @@
...
+import rhino3dm
+from pathlib import Path
+from honeybee.room import Room
+from honeybee_3dm.room import to_room
+
+
+def test_room():
+    path = str((Path.cwd()).joinpath('tests', 'test.3dm'))
+    room = to_room(path)[0]
+    assert isinstance(room, Room)
```
mostaphaRoudsari, line 11: Same comments here. You should write some better tests here. This can pass even if the translation is not what you want it to be.

mostaphaRoudsari (commented):
Hi @devngc, this version is much improved. Thank you for the fixes. I left some comments.

## module structure

I'm not sure if I understand why do we need the room and face modules. I think this library should have one module that imports a Rhino 3dm file and returns a HB Model. From there the user can use all the functionalities that built in a HB Model to do what they need.

Can you help me to understand the use case for only getting the faces or the rooms? I think once the user has the Model they can extract them as needed.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,112 @@
...
+
+    # A Layer dictionary with layer name : (Honeybee face_type, Class) structure
+    layer_to_hb_object = {
+        'roof': (face_types.roof_ceiling, Face),
+        'wall': (face_types.wall, Face),
+        'floor': (face_types.floor, Face),
+        'airwall': (face_types.air_boundary, Face),
+        'shade': (None, Shade),
+        'aperture': (None, Aperture),
+        'door': (None, Door)}
```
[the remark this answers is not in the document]
devngc: So this is a dictionary with (Honeybee face_type , Honeybee Class) structure. For the Shade, Aperture, and Door objects there are not face_types.

honeybee_3dm/face.py
devngc: This is turned into a dictionary now

honeybee_3dm/tojson.py
devngc: I looked [here](https://www.ladybug.tools/honeybee-core/docs/search.html?q=json&check_keywords=yes&area=default). I did not find a method to write hbjson from a Honeybee model. If such a method exists then I would definitely drop this module.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,112 @@
...
+from honeybee.boundarycondition import boundary_conditions
+from honeybee.typing import clean_and_id_string
+from honeybee.facetype import face_types
+from ladybug_geometry.geometry3d.face import Face3D
+
+
+def to_face(path):
+    """This function looks up a rhino3dm file, converts the objects
+    on the layer name "roof", "wall", "floor", "airwall", "shade", and  "aperture"
+    to Honeybee objects, and writes them to a json file.
```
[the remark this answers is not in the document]
devngc: Complied

honeybee_3dm/face.py
```diff
@@ -0,0 +1,112 @@
...
+    layer_to_hb_object = {
+        'roof': (face_types.roof_ceiling, Face),
+        'wall': (face_types.wall, Face),
+        'floor': (face_types.floor, Face),
+        'airwall': (face_types.air_boundary, Face),
+        'shade': (None, Shade),
+        'aperture': (None, Aperture),
+        'door': (None, Door)}
+
+    for layer in layer_to_hb_object.keys():
```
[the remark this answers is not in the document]
devngc: Complied

honeybee_3dm/face.py
```diff
@@ -0,0 +1,112 @@
...
+
+        if layer in layer_dict.keys():
```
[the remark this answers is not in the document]
devngc: Complied

honeybee_3dm/face.py
```diff
@@ -0,0 +1,112 @@
...
+            hb_face_type, hb_face_module = layer_to_hb_object[layer]
+
+            # Gathering planar geometries from the layer
```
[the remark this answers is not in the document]
devngc: A check is added.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,112 @@
...
+            rhino_faces = [object for object in rhino3dm_file.Objects
+                           if object.Attributes.LayerIndex == layer_dict[layer]]
```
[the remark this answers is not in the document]
devngc: Complied

honeybee_3dm/face.py
devngc: I have decided to remove this message as we're not concerned if there are rhino3dm objects on a layer that we don't need. We're anyway not selecting those objects.

honeybee_3dm/face.py
devngc: Removed

honeybee_3dm/face.py
devngc: the hb_face_module is different in each case. The hb_face_module it being decided based on layer name.

honeybee_3dm/room.py
devngc: Created a summary line before this.

honeybee_3dm/room.py
devngc: Complied

honeybee_3dm/room.py
devngc: Complied

honeybee_3dm/room.py
devngc: As a matter of fact, I started with to_face3d method that dealt with all these types. That method ended up having a lot of repeated code. I decided to break into smaller methods. Please note that **mesh_to_face3d** is the method that's used in other methods as well.

honeybee_3dm/togeometry.py
devngc: Honeybee [Face](https://www.ladybug.tools/honeybee-core/docs/honeybee.face.html) module takes a Ladybug Face3D object.

honeybee_3dm/togeometry.py
devngc: Honeybee [Face](https://www.ladybug.tools/honeybee-core/docs/honeybee.face.html) module takes a Ladybug Face3D object.

tests/face_test.py
devngc: Complied

devngc:
Hi @mostaphaRoudsari,
This PR is ready for another review.

honeybee_3dm/face.py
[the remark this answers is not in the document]
mostaphaRoudsari: Why don't we have a face type for those? @chriswmackey, can you confirm this?

honeybee_3dm/face.py
mostaphaRoudsari: but that is handled by the dictionary and the code stays the same. Right? In other words, you can just do it all in one check. Am I missing something here?

honeybee_3dm/grid.py
```diff
@@ -0,0 +1,47 @@
...
+    # Creating Ladybug Grid objects
+    for face in grid_face3ds:
+        try:
+            grids.append(face.mesh_grid(1, 1, 0))
+        except AssertionError:
+            pass
+
+    # Creating Honeybee Grid objects
+    hb_grids = [SensorGrid.from_mesh3d(str(i), grids[i])
+                for i in range(len(grids))]
```
mostaphaRoudsari, line 45:
```suggestion
    hb_grids = [SensorGrid.from_mesh3d(str(c), grid) for c, grid in enumerate(grids)]
```

honeybee_3dm/grid.py
```diff
@@ -0,0 +1,47 @@
...
+from honeybee_radiance.sensorgrid import SensorGrid
+from honeybee_radiance.writer import model_to_rad_folder
+
+from .room import to_room
+from .face import to_face
+from .tojson import to_json
+from .togeometry import brep2d_to_face3d
+
+
+def to_grid(path):
```
mostaphaRoudsari, line 14: I don't think this makes much sense to parse the file every single time for each object type. There should ideally be one time that you read the file and then use the loaded objects and the functions to translate them as needed.

honeybee_3dm/tojson.py
mostaphaRoudsari: Even if it is missing from the honeybee library, we should add it there and re-use it instead of adding it here.

honeybee_3dm/model.py
```diff
@@ -0,0 +1,45 @@
...
+    # Honeybee faces
+    hb_face = to_face(path)
+    # Honeybee grids
+    hb_grid = to_grid(path)
+    # Honeybee model
+    hb_model = Model.from_objects(name, hb_face)
+    # Assigning grids to Honeybee model
+    hb_model.properties.radiance.sensor_grids = hb_grid
+    # Writing Honeybee model to a Radiance folder
+    return hb_model
```
mostaphaRoudsari, line 45: This encapsulates the issue that I have been mentioning with the current structure of the modules and how they are working. You are loading the file 3 times here instead of once. I think once the file object is loaded the rest of the modules should take that as an input instead of re-loading the file from the path. Something like this:

```suggestion
    rhino3dm_file = rhino3dm.File3dm.Read(path)
    tolerance = rhino3dm_file.Settings.ModelAbsoluteTolerance

    # Honeybee faces
    hb_face = to_face(rhino3dm_file, layer_names=['faces'])
    # Honeybee grids
    hb_grid = to_grid(rhino3dm_file, layer_names=['gird'])
    # Honeybee model
    hb_model = Model.from_objects(name, hb_face)
    # Assigning grids to Honeybee model
    hb_model.properties.radiance.sensor_grids = hb_grid
    # Return a honeybee mdoel
    return hb_model
```

honeybee_3dm/model.py
```diff
@@ -0,0 +1,45 @@
...
+            model.
+
+    Returns:
+        hb_model (A list): A Honeybee models
+    """
+    # TODO create a quality check for the file path
+    # TODO create a quality check for the file name
+
+    rhino3dm_file = rhino3dm.File3dm.Read(path)
+    tolerance = rhino3dm_file.Settings.ModelAbsoluteTolerance
```
mostaphaRoudsari, line 34: Shouldn't we set the model tolerance based on this number?

As discussed today we should also set the units of the honeybee model based on the input Rhino file.

honeybee_3dm/face.py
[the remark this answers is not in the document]
devngc: [These](https://www.ladybug.tools/honeybee-core/docs/honeybee.facetype.html) are the face_types

mostaphaRoudsari (changes requested):
Hi @devngc, this is getting better and better every PR. I left a number of comments but my major request is to re-design the methods to accept the 3dm object instead of the path so you don't have to load the same file over and over.

I also think you should add helper modules to do the repetitive tasks like filtering objects by layers, and so on.

honeybee_3dm/grid.py
devngc: Complied

honeybee_3dm/tojson.py
chriswmackey: Why do we need an entire method to write a model to a JSON? It's literally 2 lines:
```
with open(hb_file, 'w') as fp:
        json.dump(model.to_dict(), fp, indent=indent)
```

chriswmackey: Looking again at the module here, this whole method is not necessary and has all of this stuff that is not relevant to writing model JSONs. Let's only focus on writing honeybee Models (not any other objects) and just use the two lines I pasted there.

honeybee_3dm/model.py
devngc: Complied

honeybee_3dm/model.py
devngc: Complied

honeybee_3dm/grid.py
devngc: Complied

honeybee_3dm/tojson.py
devngc: I created [This](https://github.com/ladybug-tools/honeybee-core/pull/274) PR. I hope I have understood the comments correctly this time.

devngc:
Hi @mostaphaRoudsari,
The pull request is ready for review except for the **tests**.

honeybee_3dm/tojson.py
devngc: Created a PR [here](https://github.com/ladybug-tools/honeybee-core/pull/277)

mostaphaRoudsari:
@devngc, I sent a PR to your fork: https://github.com/devngc/honeybee-3dm/pull/1

Let me know what you think and we can start from there. I'm going to close this PR and we can create a new fresh PR once we discuss the details under the other PR.

PR #4, opened 2020-11-13
Refactored Modules with feat(grid): Grids from mesh

devngc:


devngc:
@mostaphaRoudsari,🤞 this is ready for review 🤞

.gitignore
```diff
@@ -12,3 +12,8 @@ tox.ini
 /.vscode
 .eggs
 *.code-workspace
+
+# Rhino Backup files
+*.3dm
```
mostaphaRoudsari, line 17: We probably want to keep track of changes in 3dm files. Otherwise, the modifications to Rhino files for testing will be ignored.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,82 @@
...
+        'door': (None, Door)
+    }
+
+    for layer in rhino3dm_file.Layers:
+        if layer.Name not in layer_to_hb_object:
+            warnings.warn(
+                f'Only objects on layers {tuple(layer_to_hb_object.keys())} will be imported during the process '
+                'of importing faces.'
+            )
+            continue
```
mostaphaRoudsari, line 55: I know this is probably my mistake but we don't need continue here.

honeybee_3dm/grid.py
```diff
@@ -0,0 +1,77 @@
...
+from honeybee.typing import clean_and_id_string, clean_string
+from ladybug_geometry.geometry3d import Face3D, Mesh3D
+
+# Importing dependencies from Honeybee-3dm package
+from .togeometry import brep2d_to_face3d, mesh_to_face3d, mesh_to_mesh3d
+from .helper import filter_objects_by_layer
+
+
+# TODO: expose an option to change the target layer name
+# TODO: Grid size and offset will have to provided using the config file
```
mostaphaRoudsari, line 18: Looks like you have already addressed this TODO.

honeybee_3dm/grid.py
```diff
@@ -0,0 +1,77 @@
...
+            if isinstance(geo, rhino3dm.Brep):
+                face3d = brep2d_to_face3d(geo, tolerance)[0]
+                mesh3d = face3d.mesh_grid(grid_size, grid_size, grid_offset)
+
+            # If it's a Mesh
+            elif isinstance(geo, rhino3dm.Mesh):
+                # If it's a Mesh with only one face
+                if len(geo.Faces) == 1:
+                    face3d = mesh_to_face3d(geo)[0]
+                    mesh3d = face3d.mesh_grid(grid_size, grid_size, grid_offset)
```
mostaphaRoudsari, line 60: I think we should deal with all the meshes the same. Why are you making an exception for meshes with one face?

honeybee_3dm/grid.py
```diff
@@ -0,0 +1,77 @@
...
+                # It it's a mesh with multiple faces
+                elif len(geo.Faces) > 1:
+                    mesh3d = mesh_to_mesh3d(geo)
+        except AssertionError:
```
mostaphaRoudsari, line 64: This is a very generic error. Should we check the message?

Also if this is an error specific to Brep then we should do the try/except only for the Brep and not here.

honeybee_3dm/grid.py
```diff
@@ -0,0 +1,77 @@
...
+            warnings.warn(
+                'Grids are not created for Breps with curved edges.'
+                ' Please mesh them in Rhino and try again.')
+            # TODO: Find a way to Implement a method to create grids from a planar
+            # TODO geometry with curved edges
+            pass
```
mostaphaRoudsari, line 70: What if we use `GetMesh` method to get the render mesh in Rhino? Here is an example:

https://github.com/jesterKing/import_3dm/blob/master/import_3dm/converters/render_mesh.py#L52-L57

honeybee_3dm/grid.py
```diff
@@ -0,0 +1,77 @@
...
+
+        name = obj.Attributes.Name
+        obj_name = name.replace(' ', '_') or clean_and_id_string('grid')
```
mostaphaRoudsari, line 73: `clean_string` function in next line does find and replace white-spaces. You can remove it from here.

honeybee_3dm/model.py
```diff
@@ -0,0 +1,67 @@
...
+    hb_model = Model.from_objects(
+        name,
+        hb_faces + hb_rooms,
+        units=model_unit,
+        tolerance=model_tolerance,
+        angle_tolerance=model_angle_tolerance
+    )
+    # Assigning grids to Honeybee model
+    hb_model.properties.radiance.sensor_grids = hb_grids
+    # Writing Honeybee model to a Radiance folder
```
mostaphaRoudsari, line 66: This comment seems to be wrong.

tests/wall_test.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+from honeybee_3dm.model import import_3dm
+from ladybug_geometry.geometry3d import Point3D, Vector3D
+
+path = './tests/assets/test.3dm'
+model = import_3dm(path)
```
mostaphaRoudsari, line 6: You should probably include these lines in the test function itself.

tests/grid_test.py
```diff
@@ -0,0 +1,19 @@
...
+    # Check for User provided name
+    assert 'test-grid' in [grid.display_name for grid in model.properties.radiance.sensor_grids]
+    # Check for vertices & normal
+    # aperture = [
+    #     aperture for aperture in model.apertures if aperture.display_name == 'Window_East']
+    # assert aperture[0].vertices[0].is_equivalent(Point3D(5, 2.615265, 0.296184), 0.01)
+    # assert aperture[0].vertices[1].is_equivalent(Point3D(5, 7.105553, 0.296184), 0.01)
+    # assert aperture[0].vertices[2].is_equivalent(Point3D(5, 7.105553, 2.769152), 0.01)
+    # assert aperture[0].vertices[3].is_equivalent(Point3D(5, 2.615265, 2.769152), 0.01)
+    # assert aperture[0].normal == Vector3D(1, 0, 0)
```
mostaphaRoudsari, line 19: We should either remove these or get them fixed.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,309 @@
...
+        face = mesh.Faces[j]
+        if len(face) == 4:
+            all_verts = (pts[face[0]], pts[face[1]],
+                         pts[face[2]], pts[face[3]])
+        else:
+            all_verts = (pts[face[0]], pts[face[1]], pts[face[2]])
+
+        faces.append(Face3D(all_verts))
+
+    return faces
```
mostaphaRoudsari, line 78: Should we make a PolyFace instead? I _feel_ I already asked this and you told me the reason. Is it because Honeybee model does not support PolyFace?

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,309 @@
...
+                        "The surface has holes that touch the boundary. \
+                            It will be meshed")
+                    faces.extend(brep_to_face3d(brep))
+                else:
+                    lb_face = Face3D(boundary=boundary_pts, holes=hole_pts)
+                    faces.append(lb_face)
+    # If any of the above fails, mesh it
+    except Exception as e:
+        print(str(e) + "Face creation failed, this face will be meshed")
+        faces.extend(brep_to_face3d(brep))
```
mostaphaRoudsari, line 229: This will be confusing. We should at least separate them in outputs and let the user handle the objects with different types.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,309 @@
...
+    """
+    faces = []
+    for i in range(len(brep.Faces)):
+        mesh = brep.Faces[i].GetMesh(rhino3dm.MeshType.Any)
+        faces.extend(mesh_to_face3d(mesh))
+
+    return faces
+
+
+def brep2d_to_face3d(brep, tolerance):
```
mostaphaRoudsari, line 146: @chriswmackey, can you take a look at this function. I'm not the best person to review this.

mostaphaRoudsari (commented):
Looks good! Mostly minor comments.

honeybee_3dm/grid.py
devngc: I merely exposed on the method. There's no way for a user to provide these inputs as of now so I kept these here.

honeybee_3dm/grid.py
devngc: For the left hand side mesh (1 face) we can create grid points. For the right hand side meshes (More faces) we use mesh face centroids.
why mesh

honeybee_3dm/togeometry.py
devngc: https://www.ladybug.tools/ladybug-geometry/docs/ladybug_geometry.geometry3d.polyface.html
I believe we can write a function `solid_to_polyface3d` and replace the function [solid_to_face3d](https://github.com/devngc/honeybee-3dm/blob/master/honeybee_3dm/togeometry.py#L234). But this function is not being used as of now since we're only dealing with faces. It will be useful when we start working with Honeybee Rooms.

honeybee_3dm/grid.py
devngc: We're using the `Any` object to create mesh. Please see [here](https://github.com/devngc/honeybee-3dm/blob/master/honeybee_3dm/togeometry.py#L140)
It's not we are not able to deal with a planar brep that has curved edges. We are and we're doing that through meshing using the `Any` object similar to jesterKing. That mesh is traingulated mesh and not very useful to create grid points in my opinion. At the bottom of the image below, we have a mesh created from Rhino3dm. At the top of the image, we have a mesh created in rhino and then used for grids. I believe, you would agree that the top one is what we want and that's why I am asking the user to mesh it in rhino and try again.
why avoid curved brep

You would also ask, "how ladybug-rhino is doing this?" Well, we have methods to use the meshing parameters in rhinocommon and ladybug-rhino is using them.

While we have [meshing parameters](https://mcneel.github.io/rhino3dm/python/api/MeshingParameters.html?highlight=meshingparameter) in Rhino3dm, apparently, there's no direct way to use them. See [here](https://mcneel.github.io/rhino3dm/python/api/BrepFace.html#rhino3dm.BrepFace.GetMesh) also.

I [asked](https://discourse.mcneel.com/t/rhino3dm-meshingparameters-python/112216?u=devang_chauhan) on the forum. I am waiting for Steve Baer's response and if I don't hear from him soon then I am going to ask him why we have this whole class and no way to use it.

honeybee_3dm/grid.py
mostaphaRoudsari: I see. Good point. I might suggest to create an issue for this or add the TODO to the related module.

honeybee_3dm/grid.py
mostaphaRoudsari: I think depending on how we want to present this to user it might be confusing. I imagine similar to Ladybug we want to use the cenetriod of the Mesh as is so the user can mesh the geometries before saving them and have full control on where the sensors will end up being. We don't want to change that on your side I guess.

honeybee_3dm/grid.py
mostaphaRoudsari: I agree that the render mesh is not the best mesh for analysis but I think we should still let the study run. You are already providing a warning to user.

I don't think meshing will ever be involved in Rhino3dm. It will be kept as part of RhinoCompute.

That being said, and I have been thinking about this for some time, since this library will run outsider Rhino and IronPython we might be able to use Blender's meshing methods.

Hi @Moult, can you point us to a code sample to generate a QuadMesh using Blender library? Thanks!

honeybee_3dm/grid.py
devngc: We're not changing it. We're still doing that. My point is, if have mesh with 1 mesh face then we need go divide that into a grid of meshes first. That's why I am treating a mesh with one face separately.

honeybee_3dm/grid.py
Moult: @mostaphaRoudsari Blender has a ridiculous number of ways to remesh things, some manual, some semi automated, and some fully automated - I don't know the full history behind this but I suspect in Blender because meshes are first-class citizens in my initial implementation I left it to the user to define their topology.

But if you are looking for algorithms, here's a few options to peruse:

 - Remesh modifier https://docs.blender.org/manual/en/latest/modeling/modifiers/generate/remesh.html
 - Quadriflow remesh https://docs.blender.org/manual/en/latest/modeling/meshes/retopology.html
 - Tris to quads https://docs.blender.org/manual/en/latest/modeling/meshes/editing/face/triangles_quads.html (looking at the upper screenshot, tris to quads could further improve on the output)

devngc: Thanks @Moult, Tris to Quads looks promising. @mostaphaRoudsari, we should definitely use more of Blender.

mostaphaRoudsari: Thanks @Moult for the links. Where should we look for some sample Python code to do something like this:

image

from [here](https://docs.blender.org/manual/en/latest/modeling/modifiers/generate/remesh.html). Sorry if it is a very obvious question for Blender users but I couldn't find a useful code sample by searching `blender python remesh`. Cheers.

honeybee_3dm/grid.py
mostaphaRoudsari: >  we need go divide that into a grid of meshes first.

That's exactly my point. The input mesh should stay intact unless we remesh all the input meshes. This means that we are leaving the meshing control fully in the hands of the user when they create the meshes in Rhino. That being said if we find a better solution using Blender then we can start taking advantage of that and provide re-meshing all the input meshes as an option.

honeybee_3dm/grid.py
Moult: @mostaphaRoudsari I have a sneaking suspicion it is C++, not Python unfortunately. I'm not familiar with the C++ codebase.

See https://git.blender.org/gitweb/gitweb.cgi/blender.git
This seems like the entry point to begin code diving for the remesh modifier: https://git.blender.org/gitweb/gitweb.cgi/blender.git/blob_plain/HEAD:/source/blender/modifiers/intern/MOD_remesh.c

I think sharp mode is the one you're looking for. There is a Github mirror which makes life easier when searching: https://github.com/blender/blender/search?q=remesh

honeybee_3dm/grid.py
devngc: @mostaphaRoudsari, I propose that we talk about this. Might take 10-15 mins max. It is most likely that we both want to achieve the same thing.

devngc: I think I got your point. You want that we don't divide a one face mesh into grids and leave that to the users. I will make this change.

honeybee_3dm/grid.py
devngc: Created [here](https://github.com/ladybug-tools/honeybee-3dm/issues/5)

honeybee_3dm/togeometry.py
devngc: @mostaphaRoudsari,  In this method we have two checks [here](https://github.com/devngc/honeybee-3dm/blob/master/honeybee_3dm/togeometry.py#L159) and [here](https://github.com/devngc/honeybee-3dm/blob/master/honeybee_3dm/togeometry.py#L216). I have kept this exception to catch any other scenario for which we have not created a check yet. I would remove this message and let the user see the error message instead.

Do you also want me to report the object to the user? I can print the object ID and then a user can use `SelID` in rhino to quickly select the object.

mostaphaRoudsari: I follow your logic and I see what are you trying to do but it will not address my initial conerns. This is more of a "code smell" than a bug. The structure of the code is not clean enough. If you are expecting an error it should be checked where the error happens.

```python
try:
    # small code here
expect ValueError:
   # handle the error
else:
   # no error - do what needs to be done
```

You can have multiple of these instead of one. In the current code, so many things can go wrong and a single try/expect is not necessarily the best practice. I'm not familiar enough with the SDK to comment on details. That's why I asked for @chriswmackey's help here.

honeybee_3dm/grid.py
mostaphaRoudsari: I'm resolving this for now. @devngc, we should probably open an issue and link this conversation to address it at some point in the near future. Thanks @Moult!

devngc: Created a new issue

honeybee_3dm/grid.py
devngc: @mostaphaRoudsari,
I have addressed it.

honeybee_3dm/togeometry.py
devngc: @mostaphaRoudsari ,
I changed this and made the exception handling specific.

mostaphaRoudsari (approved):
@devngc, thank you for all the improvements. It looks good to me.

I would wait for @chriswmackey for another day to see if he can review that single method. Other than that it's good to be merged IMO.

honeybee_3dm/face.py
```diff
@@ -0,0 +1,81 @@
...
+        tolerance: A number for model tolerance. By default the tolerance is set to
+            the ModelAbsoluteTolerance value in input 3DM file.
+
+    Returns:
+        A list of Honeybee faces.
+
+    """
+    hb_faces = []
+
+    # TODO: Add an input to customize layer names
```
chriswmackey, line 37: Ah. This will be so nice!

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+        except AttributeError:
+            warnings.warn(
+                'Non-planar Breps are not handled by Honeybee. They are ignored.'
+            )
+            return faces
+        else:
+            # In the list of the Polylines, if there's only one polyline then
+            # the face has no holes
+            if len(sorted_polylines) == 1:
+                if len(mesh_pts) == 4 or len(mesh_pts) == 3:
```
chriswmackey, line 202: Couldn't you just do this check at the very start?  If there are only 3 or 4 vertices, then there's no need to do this whole crazy thing with the polyline edges, etc.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+
+        # Getting all vertices of the face
+        mesh = brep.Faces[0].GetMesh(rhino3dm.MeshType.Any)
+        try:
+            mesh_pts = [to_point3d(mesh.Vertices[i])
+                        for i in range(len(mesh.Vertices))]
+        except AttributeError:
+            warnings.warn(
+                'Non-planar Breps are not handled by Honeybee. They are ignored.'
+            )
```
chriswmackey, line 196: I don't understand why you would get an AttributeError for a non-planar Brep face. Don't all Breps have a render mesh?

honeybee_3dm/togeometry.py
chriswmackey: Or maybe you could check if the number of Brep edges is 3 or 4.   Also, a hole is impossible for any shape with 5 edges or less so maybe you can save yourself a lot of extra checks in these cases by just ignoring the search for holes in these cases.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+            return faces
+        else:
+            # In the list of the Polylines, if there's only one polyline then
+            # the face has no holes
+            if len(sorted_polylines) == 1:
+                if len(mesh_pts) == 4 or len(mesh_pts) == 3:
+                    faces.extend(mesh_to_face3d(mesh))
+                else:
+                    boundary_pts = remove_dup_vertices(
+                        sorted_polylines[0].vertices, tolerance)
```
chriswmackey, line 206: Why do you need to do this here?  The `Face3D` method has it's own method for removing duplicate vertices which people can optionally do afterwards if its needed for a simulation.   For example, Radiance doesn't care about duplicate vertices but EnergyPlus does (so all of the to_idf methods and to_osm methods remove them). Also, you are missing the case here where a 4-vertex triangular face has a duplicate vertex.  So it's not like your code here is actually reliable anyway.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+                    lb_face = Face3D(boundary=boundary_pts)
+                    faces.append(lb_face)
+
+            # In the list of the Polylines, if there's more than one polyline then
+            # the face has hole / holes
+            elif len(sorted_polylines) > 1:
+                boundary_pts = remove_dup_vertices(
+                    sorted_polylines[0].vertices, tolerance)
+                hole_pts = [remove_dup_vertices(polyline.vertices, tolerance)
+                            for polyline in sorted_polylines[1:]]
```
chriswmackey, line 216: Again, you don't need to remove the duplicate vertices here.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+                    pts for pts_lst in hole_pts for pts in pts_lst]
+                hole_pts_on_boundary = [
+                    pts for pts in total_hole_pts if pts in boundary_pts]
+                # * Check 02 - If any of the hole is touching the boundary of the face
+                # then mesh it
+                if len(hole_pts_on_boundary) > 0:
+                    warnings.warn(
+                        "The surface has holes that touch the boundary. \
+                            It will be meshed")
+                    faces.extend(brep_to_face3d(brep))
```
chriswmackey, line 228: Meshing here seems acceptable to me.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+                else:
+                    lb_face = Face3D(boundary=boundary_pts, holes=hole_pts)
+                    faces.append(lb_face)
+
+    return faces
+
+
+def solid_to_face3d(brep):
+    """Create a Ladybug Face3D object from a rhino3dm Brep.
```
chriswmackey, line 237:
```suggestion
    """Get a list of Ladybug Face3D objects from a solid rhino3dm Brep.
```

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+        A list of Ladybug Face3D objects.
+    """
+    faces = []
+    mesh = extrusion.GetMesh(rhino3dm.MeshType.Any)
+    faces.extend(mesh_to_face3d(mesh))
+
+    return faces
+
+
+def to_face3d(obj, *, tolerance, raise_exception=True):
```
chriswmackey, line 272: What is this second input argument here? I don't see it in the input Args.

requirements.txt
```diff
@@ -1,2 +1,2 @@
-honeybee-core==1.39.4
+lbt-honeybee==0.2.221
```
chriswmackey, line 1: You should update this before merging.

chriswmackey (changes requested):
Thanks @devngc . It generally looks good. @mostaphaRoudsari was right that the `brep2d_to_face3d` method needs some work.  But almost everything else is looking great.

As a side note, it would be good if you could add more unit tests before merging.

honeybee_3dm/face.py
devngc: Yes. We will have to implement this for the users' flexibility.

honeybee_3dm/helper.py
```diff
@@ -0,0 +1,83 @@
...
+    try:
+        file_unit = file_3dm.Settings.ModelUnitSystem
+    except AttributeError:
+        raise TypeError(f'Expected a Rhino 3dm file object not {type(file_3dm)}')
+
+    unit = str(file_unit).split('.')[-1]
+
+    if unit not in UNITS:
+        raise ValueError(
+            f'{unit} is not currently supported. Supported units are {UNITS}.'
```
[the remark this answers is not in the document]
devngc: In python 3 you can. And this is introduced by Mostapha.

honeybee_3dm/model.py
```diff
@@ -0,0 +1,67 @@
...
+import rhino3dm
+
+# Importing core Honeybee dependency
+from honeybee.model import Model
+from honeybee.typing import clean_string
+
+# Importing dependencies from Honeybee-3dm package
+from .room import import_rooms
+from .face import import_faces
+from .grid import import_grids
```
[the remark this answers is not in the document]
devngc: Yes. We have a lot more to do here. Mostapha wanted me to create a PR so that you both can review it. And I can see that this process is very important.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+        color_by_face: Bool. True will derive colors from mesh faces
+
+    Returns:
+        A tuple of mesh faces and mesh colors
+    """
+
+    colors = None
+    lb_faces = []
+    for i in range(len(mesh.Faces)):
+        face = mesh.Faces[i]
```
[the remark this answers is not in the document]
devngc: What you recommend is the pythonic way to do it and I would like to do it. However, when I do it here, for some reason it creates an almost unending list of mesh faces and I have to terminate the process. I showed this to Mostapha as well. So I have to use these studs which give the correct number of faces.

honeybee_3dm/grid.py
```diff
@@ -0,0 +1,75 @@
...
+            try:
+                mesh3d = face3d.mesh_grid(grid_size, grid_size, grid_offset)
+            except AssertionError as e:
+                warnings.warn(
+                    f'{e} Breps with curved edges are not supported'
+                    ' for grids. Mesh them in Rhino and try again.'
+                )
+                continue
+            # TODO: Find a way to Implement a method to create grids from a planar
+            # TODO geometry with curved edges
```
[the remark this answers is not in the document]
devngc: I have tested this. We are able to create Face3D from a Brep with curved edge. The following image shows the Honeybee Floor objects on a Brep with curved edges.
meshes
So the error happens on the next line where I try to mesh them into grids.

honeybee_3dm/model.py
```diff
@@ -0,0 +1,67 @@
...
+    model_unit = get_unit_system(rhino3dm_file)
+
+    # Honeybee faces
+    hb_faces = import_faces(rhino3dm_file, model_tolerance)
+    # Honeybee_rooms
+    hb_rooms = import_rooms(rhino3dm_file, model_tolerance)
+    # Honeybee grids
+    hb_grids = import_grids(rhino3dm_file, model_tolerance)
+    # Honeybee model
+    hb_model = Model.from_objects(
```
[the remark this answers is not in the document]
devngc: If it improves speed, we are doing it. I did not have faces, shades, apertures, and doors in separate lists but I fixed that and now am initializing the model object directly.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+    """Ladybug Mesh3D from Rhino Mesh.
+
+    Args:
+        mesh: Rhino3dm mesh
+        color_by_face: Bool. True will derive colors from mesh faces
+
+    Returns:
+        A Ladybug Mesh3D object
+    """
+    lb_verts = tuple(to_point3d(mesh.Vertices[i]) for i in range(len(mesh.Vertices)))
```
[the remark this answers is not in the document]
devngc: The issue here is similar to what I mentioned above. The pythonic way gives the wrong number of vertices.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+    return faces
+
+
+def brep2d_to_face3d(brep, tolerance):
+    """Create a Ladybug Face3D object from a rhino3dm Brep.
+
+    This function is used in translating rhino3dm planar Brep to Ladybug Face3D objects.
+
+    Args:
+        brep: A rhino3dm Brep.
```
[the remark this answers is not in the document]
devngc: Complied

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+
+    Returns:
+        A list of Ladybug Face3D objects.
+    """
+    faces = []
+    # * Check 01 - If any of the edge is an arc
+    lines = []
+    not_line = 0
+    for i in range(len(brep.Edges)):
+        if brep.Edges[i].IsLinear(tolerance) is False:
```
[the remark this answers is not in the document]
devngc: Complied

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+    """Create a Ladybug Face3D object from a rhino3dm Brep.
+
+    This function is used in translating rhino3dm planar Brep to Ladybug Face3D objects.
+
+    Args:
+        brep: A rhino3dm Brep.
+
+    Returns:
+        A list of Ladybug Face3D objects.
+    """
```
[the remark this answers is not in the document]
devngc: Implemented a check for planarity of the Brep

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+    lines = []
+    not_line = 0
+    for i in range(len(brep.Edges)):
+        if brep.Edges[i].IsLinear(tolerance) is False:
+            not_line += 1
+    # If there's an arc in one of the edges, mesh it
+    if not_line > 0:
+        warnings.warn(
+            "The surface has curved edges. It will be meshed.")
+        faces.extend(brep_to_face3d(brep))
```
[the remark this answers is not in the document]
devngc: I would really like to look into this. for now, I am creating an [issue](https://github.com/ladybug-tools/honeybee-3dm/issues/7) so I can address this later.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+        for i in range(len(brep.Edges)):
+            start_pt = to_point3d(brep.Edges[i].PointAtStart)
+            end_pt = to_point3d(brep.Edges[i].PointAtEnd)
+            line = LineSegment3D.from_end_points(start_pt, end_pt)
+            lines.append(line)
+
+        # Create Ladybug Polylines from the lines
+        polylines = Polyline3D.join_segments(lines, tolerance)
+        # Sort the Ladybug Polylines based on length
+        # The longest polyline belongs to the boundary of the face
```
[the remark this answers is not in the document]
devngc: Revised

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+    Returns:
+         A list of Ladybug Point3D objects with duplicate points removed.
+
+    """
+    return [pt for i, pt in enumerate(vertices)
+            if not pt.is_equivalent(vertices[i - 1], tolerance)]
+
+
+def mesh_to_face3d(mesh):
+    """Create a Ladybug Face3D object out of a rhino3dm Mesh.
```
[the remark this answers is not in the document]
devngc: Revised

honeybee_3dm/togeometry.py
mostaphaRoudsari: The bare `*` forces the use of named arguments after it. It's for good practice to decrease the chance of making mistake because of the order of inputs. See here: https://stackoverflow.com/questions/14301967/bare-asterisk-in-function-arguments

Calling this function as `to_face3d(obj, 0.1)` will throw an error. The arguments after * should be called by name: `to_face3d(obj, tolerance=0.1)`

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,311 @@
...
+            end_pt = to_point3d(brep.Edges[i].PointAtEnd)
+            line = LineSegment3D.from_end_points(start_pt, end_pt)
+            lines.append(line)
+
+        # Create Ladybug Polylines from the lines
+        polylines = Polyline3D.join_segments(lines, tolerance)
+        # Sort the Ladybug Polylines based on length
+        # The longest polyline belongs to the boundary of the face
+        # The rest of the polylines belong to the holes in the face
+        sorted_polylines = sorted(polylines, key=lambda polyline: polyline.length)
```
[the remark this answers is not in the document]
devngc: Revised

honeybee_3dm/togeometry.py
devngc: I have revised as per the comment above. I don't look for holes once there's only one Polyline.

honeybee_3dm/togeometry.py
devngc: The vertices extracted from mesh have duplicates and unless I remove them I cannot create a Face3D. That's why I had to create that helper function to remove duplicate vertices.

honeybee_3dm/togeometry.py
devngc: Below are two images. The first one is when duplicate vertices are removed. The second one is when they are not removed.
duplicate vertices removed
duplicate vertices not removed

honeybee_3dm/togeometry.py
devngc: Revised

requirements.txt
devngc: Following is what I have tin requirements;
lbt-honeybee==0.2.221
rhino3dm==0.13.0
I am not sure if I need to update something here.

honeybee_3dm/togeometry.py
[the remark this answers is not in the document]
mostaphaRoudsari: I can confirm that there was something wrong with how the iterator was working. Maybe we should add a comment to the code itself and explain why it's done like this for future reference.

devngc:
@chriswmackey, I have addressed comments and I pushed the new changes. I am mostly working on expanding the test now.

devngc:
@mostaphaRoudsari, why do we not see Travis building this PR?

honeybee_3dm/togeometry.py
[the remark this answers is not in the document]
chriswmackey: Thanks for confirming. I guess it's just a but in Rhino3dm then.

honeybee_3dm/grid.py
```diff
@@ -0,0 +1,83 @@
...
+            if check_planarity(geo):
+                face3d = brep2d_to_face3d(geo, tolerance)[0]
+
+                # Check if the Brep has curved edges
+                try:
+                    mesh3d = face3d.mesh_grid(grid_size, grid_size, grid_offset)
+                except AssertionError:
+                    warnings.warn(
+                        'Breps with curved edges are not supported'
+                        ' for grids as of now. Please Mesh them in Rhino and try again.'
```
chriswmackey, line 62: This still doesn't seem right. What is the `AssertionError` that you get from `ladybug_geometry` when you have a face with a curved edge?  Are you sure that you always get an `AssertionError` for this case?

Also, if the Brep is planar but with curved edges, can't you just mesh it into a bunch of Face3D triangles, join the triangles into a Polyface3D, extract the naked edges and join these edges into a single Face3D?

honeybee_3dm/grid.py
```diff
@@ -0,0 +1,83 @@
...
+                    )
+                    continue
+                # TODO: Find a way to Implement a method to create grids from a planar
+                # TODO geometry with curved edges
```
chriswmackey, line 66: Actually, this comment makes it clear that you're thinking of implementing the method I listed with naked edges.  Let's just make sure that the `AssertionError` is reliable.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,330 @@
...
+
+    Returns:
+        A Ladybug Mesh3D object
+    """
+    lb_verts = tuple(to_point3d(mesh.Vertices[i]) for i in range(len(mesh.Vertices)))
+    lb_faces, colors = extract_mesh_faces_colors(mesh, color_by_face)
+    return Mesh3D(lb_verts, lb_faces, colors)
+
+
+def brep_to_face3d(brep):
```
chriswmackey, line 129: The name for this method doesn't really seem appropriate. Technically, you're converting it to a list of triangulated Face3Ds or meshed Face3Ds.  What about `brep_to_meshed_face3d`?  Or, if you would rather not change the name, at least update the docstring to make it clear that this method uses the meshing pathway.

honeybee_3dm/togeometry.py
chriswmackey: Interesting. Thank you for explaining, @mostaphaRoudsari . That's pretty nifty.

requirements.txt
chriswmackey: The current version of `lbt-honeybee` is 0.2.249:
https://github.com/ladybug-tools/lbt-honeybee

The version gets incremented with every change that we merge for honeybee or ladybug so you should probably check it right before your merge.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,330 @@
...
+        A list of Ladybug Face3D objects.
+    """
+    faces = []
+    # * Check 01 - If any of the edge is an arc
+    lines = []
+
+    # Getting all vertices of the face
+    mesh = brep.Faces[0].GetMesh(rhino3dm.MeshType.Any)
+    mesh_pts = [to_point3d(mesh.Vertices[i])
+                for i in range(len(mesh.Vertices))]
```
chriswmackey, line 179: It looks like you don't end up using these points at all so I would just get rid of this variable.  You can replace it with a check of `len(mesh.Vertices)` later in the script.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,330 @@
...
+            lines.append(line)
+
+        # Create Ladybug Polylines from the lines
+        polylines = Polyline3D.join_segments(lines, tolerance)
+
+        # In the list of the Polylines, if there's only one polyline then
+        # the face has no holes
+        if len(polylines) == 1:
+            if len(mesh_pts) == 4 or len(mesh_pts) == 3:
+                faces.extend(mesh_to_face3d(mesh))
```
chriswmackey, line 207: You should move this check all of the way to the beginning of the `else:` above.  If there are only 3 or 4 vertices, then there's no need to create a join the edges at all.  Not creating the edge objects is going to save a lot of time and memory in the case that the Rhino model has a lot of rectangular surfaces.

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,330 @@
...
+            else:
+                boundary_pts = remove_dup_vertices(
+                    polylines[0].vertices, tolerance)
```
chriswmackey, line 210: Can you put a comment here to explain why you need to remove the vertices for this case but not for any of the others?

chriswmackey (commented):
We're getting close, @devngc .  I just have a few comments left.

honeybee_3dm/togeometry.py
chriswmackey: That's a Rhino display issue. I think you should just not remove vertices in your code and you can have a separate function that runs the `ladybug_geometry` `reomve_colinear_vertices` method if people want to have ALL of the duplicate vertices removed:

https://www.ladybug.tools/ladybug-geometry/docs/ladybug_geometry.geometry3d.face.html#ladybug_geometry.geometry3d.face.Face3D.remove_colinear_vertices

honeybee_3dm/togeometry.py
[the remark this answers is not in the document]
devngc: I already complied with this comment. Can we consider this resolved?

honeybee_3dm/togeometry.py
devngc: This is not relevant anymore. Can we consider this resolved?

honeybee_3dm/togeometry.py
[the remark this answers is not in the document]
devngc: I already complied with this comment. Can we consider this resolved?

honeybee_3dm/togeometry.py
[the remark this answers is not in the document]
devngc: I already complied with this comment. Can we consider this resolved?

honeybee_3dm/togeometry.py
devngc: I will try to once again demonstrate the issue that happens when I don't remove duplicate vertices.
This is a surface in Rhino with its vertices visible
vertices
In the image below, you can see the points for the boundary and the holes. Please also see the duplicate vertices at the end of each tuple of points.
points
When I allow the duplicate vertices the outcome of the Honeybee face is the following;
Honeybee Floor

If this is a rhino issue then I need a way to fix it. Do I not? Removing duplicate vertices is the way I found. I am not aware of a better way but I can look for it if removing vertices is not acceptable.

Removal of duplicate vertices happens before I create a Face3D object. I am not sure how do you want me to use this `remove_colinear_vertices ` method for a Face3D object before creating a Face3D.

I have a feeling that I am missing something here.

honeybee_3dm/grid.py
devngc: I implemented what you recommended here and I love it.

requirements.txt
devngc: Okay. I will keep this in mind when we merge.

honeybee_3dm/grid.py
devngc: I implemented the method.

honeybee_3dm/togeometry.py
devngc: Complied. I renamed the function.

honeybee_3dm/togeometry.py
devngc: Complied.

honeybee_3dm/togeometry.py
devngc: Complied.

honeybee_3dm/togeometry.py
devngc: Complied. Added comments at both the places where duplicate vertices are removed.

honeybee_3dm/grid.py
```diff
@@ -0,0 +1,69 @@
...
+
+    hb_grids = []
+
+    for obj in grid_objs:
+        geo = obj.Geometry
+        # If it's a Brep
+        if isinstance(geo, rhino3dm.Brep):
+            # Check if the Brep is planar
+            if check_planarity(geo):
+                face3d = brep_to_face3d(geo, tolerance)[0]
```
chriswmackey, line 52: Shouldn't you be creating a mesh grid for all of the `Face3Ds` that this method returns?  You can create multiple Meshes and then join them using the method here:
https://www.ladybug.tools/ladybug-geometry/docs/ladybug_geometry.geometry3d.mesh.html#ladybug_geometry.geometry3d.mesh.Mesh3D.join_meshes

honeybee_3dm/togeometry.py
```diff
@@ -1,34 +1,364 @@
...
+    Args:
+        brep: A rhino3dm Brep.
+        tolerance: A rhino3dm tolerance object. Tolerance set in the rhino file.
+
+    Returns:
+        A list of Ladybug Face3D objects.
+    """
+    faces = []
+    # Getting all vertices of the face
+    mesh = brep.Faces[0].GetMesh(rhino3dm.MeshType.Any)
```
chriswmackey, line 204: Does this mean that Brep geometry must always be exploded in order to use your library here?  It's just a suggestion but maybe you should just iterate through the faces and translate each one to a `Face3D`.

chriswmackey (approved):
This generally looks good, @devngc .  I just have a cupule of suggestions that you should feel free to take or leave before you merge.

honeybee_3dm/togeometry.py
devngc: Yes, in this library, we always create a Face3D for each of the face of a rhino3dm Brep. I am already doing that. I am guessing that this suggestion stems from you observation that I am only using one face of the brep here. `brep.Faces[0]`. PLease note that this method is only employed for planar Breps, so in most cases there's only one face. For the rest of the cases, I run the Brep through the `brep_to_meshed_face3d` method where the brep is converted into a mesh > The mesh faces are converted into Face3ds > A plant polysurface is created from face3ds > A combined Face3d is created from the polyface3d.

honeybee_3dm/grid.py
devngc: I revised the brep_to_face3d method to return a single Ladybug Face3D object. I realize it was creating confusion as to why I am using just the first Face3d object.

PR #22, opened 2020-11-25
Layer flexibility

### LINKED ISSUE #20: Layer structure control
Make sure that a user can tie an existing layer to once of the default Honeybee layers through the config file.

devngc:
resolves #20


honeybee_3dm/face.py
```diff
@@ -27,7 +27,10 @@ def import_faces(rhino3dm_file, tolerance=None):
         rhino3dm_file: A Rhino3DM file object.
         tolerance: A number for model tolerance. By default the tolerance is set to
             the ModelAbsoluteTolerance value in input 3DM file.
-
+        visibility: Bool. If set to False then the objects on an "off"
+            layer in Rhino3dm will also be imported. Defaults to True.
+        config: A dictionary. The config file as a dictionary object.
+            Defaults to None.
```
mostaphaRoudsari, line 33: This is very confusing. Is it supposed to be a dictionary or a file?

honeybee_3dm/face.py
```diff
@@ -36,63 +39,115 @@ def import_faces(rhino3dm_file, tolerance=None):
...
-            for face_obj in lb_faces:
-                # TODO: Double check with Chris if this naming works for energy models.
-                # if name is assigned by user use the same name for all the sub faces
-                # otherwise generate a randome name based on the layer name.
-                obj_name = name or clean_and_id_string(layer)
-                args = [clean_string(obj_name), face_obj]
-                if hb_face_type:
-                    args.append(hb_face_type)
-                    hb_face = hb_face_module(*args)
+    config = config
```
mostaphaRoudsari, line 42: why do you need this?

honeybee_3dm/face.py
```diff
@@ -36,63 +39,115 @@ def import_faces(rhino3dm_file, tolerance=None):
...
+    # If config file is provided and any Rhino layer is assigned
+    if config and config['HB_layers'].values():
```
mostaphaRoudsari, line 44: You need to check that `HB_layers` key exist in a different line before trying to get the values here. This will raise a non explicit error if the provided dictionary is missing the key.

honeybee_3dm/face.py
```diff
@@ -36,63 +39,115 @@ def import_faces(rhino3dm_file, tolerance=None):
...
+        # Map rhino layers to HB_layers based on the config file
+        hb_layer_from_layer = {config['HB_layers'][key]: key for key in
+            config['HB_layers']}
```
mostaphaRoudsari, line 47:
```suggestion
        hb_layer_from_layer = {value: key for key, value in config['HB_layers'].items()}
```

honeybee_3dm/face.py
```diff
@@ -36,63 +39,115 @@ def import_faces(rhino3dm_file, tolerance=None):
...
+            HB_layers.floor.value: (face_types.floor, Face),
+            HB_layers.airwall.value: (face_types.air_boundary, Face),
+            HB_layers.shade.value: (None, Shade),
+            HB_layers.aperture.value: (None, Aperture),
+            HB_layers.door.value: (None, Door)
+        }
+        # A dictionary with child layer : parent layer structure
+        child_parent = child_parent_dict(rhino3dm_file)
+
+        for layer in rhino3dm_file.Layers:
```
mostaphaRoudsari, line 61: This is a very long for loop and I couldn't really follow the logic. I suggest to break down the code in smaller reusable functions/pieces.

honeybee_3dm/grid.py
```diff
@@ -27,43 +28,48 @@ def import_grids(rhino3dm_file, tolerance, grid_size=1, grid_offset=0):
...
-    except ValueError:
-        # no layer named grid
-        return []
-
     hb_grids = []
+    config = config
+    # If config file is provided and any Rhino layer is tied to HB_grid layer
+    if config and config['HB_layers']['HB_grid']:
+        grid_objs = objects_on_parent_child(rhino3dm_file, 
+            config['HB_layers']['HB_grid'], visibility=visibility)
```
mostaphaRoudsari, line 43: You can avoid having several nested layers of code. It is really hard to follow the logic of 4-5 if/else one after another.

```python
if not config or 'HB_grid' not in config['HB_layers']:
   return hb_grids

# now start the new check here
```

honeybee_3dm/grid.py
```diff
@@ -27,43 +28,48 @@ def import_grids(rhino3dm_file, tolerance, grid_size=1, grid_offset=0):
...
+                # Check if the Brep is planar
+                if check_planarity(geo):
+                    try:
+                        mesh3d = brep_to_face3d(geo, tolerance).mesh_grid(grid_size,
+                            grid_size, grid_offset)
+                    except AttributeError:
+                        warnings.warn(
+                            'Please turn on the shaded mode in rhino, save the file,'
+                        ' and try again.'
+                        )
```
mostaphaRoudsari, line 58: If this means that you are unable to import critical data it should raise an error and not just a warning. User can ignore the warning and not notice that the file hasn't been exported correctly.

honeybee_3dm/helper.py
```diff
@@ -1,4 +1,41 @@
...
+
+@unique
+class HB_layers(Enum):
+    """Default Honeybee layers.
+    
+    This class hosts all the default Honeybee layers in
+    layer key : layer name structure
+
+    Args:
+        Enum: An Enum object.
```
mostaphaRoudsari, line 13: I'm not sure if this is really helpful. Also Enum is the baseclass and not an argument.

honeybee_3dm/helper.py
```diff
@@ -1,4 +1,41 @@
...
+
+    Args:
+        path: A text string for the path to the config.json.
+
+    Returns:
+        A dictionary for the config.json
+    """
+    with open(path) as fh:
+        config = json.load(fh)
+        return config
```
mostaphaRoudsari, line 38: It's ok to keep this but honestly I think just writing the code for loading the JSON data is just as good.

honeybee_3dm/helper.py
```diff
@@ -27,63 +64,163 @@ def get_unit_system(file_3dm):
...
+        A bool. True if the layer is found tied to one of the HB_layers in the config
+            file.
     """
-    layer_index = [layer.Index for layer in file_3dm.Layers if layer.Name == layer_name]
-    if not layer_index:
-        raise ValueError(f'Find no layer named "{layer_name}"')
+    if layer_name in config['HB_layers'].values() \
+        and layer_name != config['HB_layers']['HB_grid'] \
+        and layer_name != config['HB_layers']['HB_room'] \
+        and layer_name != config['HB_layers']['HB_view']:
```
mostaphaRoudsari, line 81: This will create strange errors when the user doesn't provide a full dictionaries with all the keys that you are expecting. Do you have a function to check that and create None values for missing keys?

honeybee_3dm/helper.py
```diff
@@ -27,63 +64,163 @@ def get_unit_system(file_3dm):
...
+        return True
+    else:
+        return False
 
-    return filter_objects_by_layer_index(file_3dm, layer_index[0])
 
+def check_layer_not_in_hb_layers(config, layer_name, child_parent):
```
mostaphaRoudsari, line 87: The name suggests something different than the function do. Is it `check_layer_pair_in_hb_layers`?

tests/assets/config.json
```diff
@@ -0,0 +1,15 @@
+
+{
+    "HB_layers" :{
```
mostaphaRoudsari, line 3: Why do you need an extra key here? Can't we just have a flattened structure if you are only using this file for layers and remove the "HB_layers" key?

```
{       "HB_wall": "",
        "HB_grid": "grid",
        "HB_room": "",
        "HB_floor": " ",
        "HB_roof": " ",
        "HB_airwall": " ",
        "HB_view": " ",
        "HB_aperture": "window",
        "HB_shade": "shade",
        "HB_door": "door"
    }
```

mostaphaRoudsari (changes requested):
Hi @devngc, I didn't comment on every detail but left some high level comments. I think both the code and the suggested layer structure can be simplified. There are just way too many checks for a single action in a single function. I suggest to refactor the code to be easier to understand.

PR #23, opened 2020-11-25
Grid controls

### LINKED ISSUE #5: Grid controls for Users
Currently, the grid controls are exposed in the [grid module](https://github.com/devngc/honeybee-3dm/blob/master/honeybee_3dm/grid.py#L21). However, there's no way for a user to use these controls as of now. 
The capability to provide grid controls will be handled through the config file.

devngc:
resolves #5


mostaphaRoudsari (changes requested):
The config file structure needs some work. Right now it is really rigid and complicated. It has also resulted in so many checks in code. What if we change the structure of the config file and make it more explicit.

The information that you really want to include are assigning materials to different layers in Rhino and also setting the grid size for grid layers. 

```python
{
    'rad_materials':  {'layer_name' : 'material_name', ... },
    'grid_settings': {'layer_name': { 'grid_size': 0.6, 'grid_offset': 0.76}, ...},
    ...
}
```

This structure makes your checks much easier for `import` functions and is also extendable with additional keys for energy_materials as so on. It also doesn't need to edit the material names inside the Rhino file.

I suggest to document your ideas for such developments as a proposal on Basecamp before implementation. That will ensure that we discuss details like this before jumping into the development.

PR #24, opened 2020-11-27
Radiance Material file integration

devngc:
This PR adds the capability to use radiance modifiers from a .mat file.


mostaphaRoudsari (changes requested):
See my comment on other PRs. I think we should discuss the purpose and the structure of the config file before adding any new code.

Also if all the PRs are built on top of each other we should just send a single PR. There is no point of 4 separate PRs if each one includes all the changes in the previous PRs.

PR #33, opened 2020-12-07
Config Schema generation capability, rhino style layer visibility, refactoring

### LINKED ISSUE #5: Grid controls for Users

### LINKED ISSUE #20: Layer structure control

devngc:
This PR resolves #20 and resolves #5.

This PR has all the modules refactored to work with the latest config file structure.

The PR also addressed the major comments from past PRs [22](https://github.com/ladybug-tools/honeybee-3dm/pull/22) and [10](https://github.com/ladybug-tools/honeybee-3dm/pull/10) The summary of major comments is following;

1. Create a more extendable config file structure
2. Refactor codebase to avoid nested functions
3. Rename the argument name visibility to layer_visibility (This is not relevant anymore)
4. Raise an error if an object in rhino is not saved with shaded mesh
5. Implement quality check for the config file 

The rest of the comments on the above mentioned PRs were irrelevant due to refactoring of the codebase.


honeybee_3dm/config.py
```diff
@@ -0,0 +1,432 @@
...
+        Boolean value of True if no error is raised.
+    """
+    layers = [layer.Name for layer in file_3dm.Layers]
+    layer_check = [True for layer in config['layers'] if layer in layers]
+    if len(config['layers']) != layer_check.count(True):
+        raise KeyError(
+            'Only layer names from the Rhino file are allowed in the config file.'
+        )
+    else:
+        return True
```
mostaphaRoudsari, line 166: Am I right that you are trying to make sure all the layers in config file are also in Rhino? In that case can't we just check that?

```suggestion
    layer_check = [layer for layer in config['layers'] if layer not in layers]
    if layer_check:
        raise KeyError(
            'Only layer names from the Rhino file are allowed in the config file. '
            f'Found invalid layer names: {layer_check}'
        )
    return True
```

mostaphaRoudsari: Also what about child layers? Should we support `parent_layer::child_layer' for layer names?

honeybee_3dm/config.py
```diff
@@ -0,0 +1,432 @@
...
+
+
+def check_rad(config):
```
mostaphaRoudsari, line 169: Since this check only has one input which is config - I think this is a better fit to be added to Config as a validator.

honeybee_3dm/config.py
```diff
@@ -0,0 +1,432 @@
...
+    Config.parse_file(config_path)
+
+    # Get a config dictionary from the config file
+    config = read_json(config_path)
+
+    # Validate config with external binaries
+    if check_layers(file_3dm, config) and check_rad(config):
+        return config
+    else:
+        None
```
mostaphaRoudsari, line 246: ?

honeybee_3dm/config.py
```diff
@@ -0,0 +1,432 @@
...
+
+def check_config(file_3dm, config_path):
+    """Validates the config file and returns it in the form of a dictionary.
+
+    Args:
+        file_3dm: A rhino3dm file object.
+        config_path: A text string for the path to the config file.
+
+    Returns:
+        Config dictionary or None if any of the checks fails.
```
mostaphaRoudsari, line 234: Why don't you just work with the `Config` object instead of exporting it as a dictionary?

honeybee_3dm/config.py
```diff
@@ -0,0 +1,432 @@
...
+
+    # Validate config with external binaries
+    if check_layers(file_3dm, config) and check_rad(config):
+        return config
+    else:
+        None
+
+
+
+def child_layer_control(config, layer_name):
```
mostaphaRoudsari, line 250: This can be a method added to the Config object.

honeybee_3dm/config.py
```diff
@@ -0,0 +1,432 @@
...
+    if 'radiance_material' in config['layers'][layer_name]:
+        radiance_modifiers = mat_to_dict(config['sources']['radiance_material'])
+        hb_face.properties.radiance.modifier = radiance_modifiers[config
+            ['layers'][layer_name]['radiance_material']]
+        return hb_face
+    else:
+        return hb_face
+
+
+def face3d_to_hb_face_with_rad(config, face_obj, name, layer_name):
```
mostaphaRoudsari, line 349: You should probably move all these methods to a different module and keep this module limited to Config.

mostaphaRoudsari (changes requested):
I left my comments on the config file. Let's focus on the config object first and then we can review the rest of the modules. The class itself looks good. I mostly added suggestions for descriptions.

Also I'm not sure why would we want to convert the Config object to a dictionary? Can't we just use it as is?

devngc:
@mostaphaRoudsari, thanks for the review and comments. Config is a dictionary object itself. I am not recreating it.

mostaphaRoudsari:
I meant you should parse it using the Config class and then use the object instead of the dictionary. Otherwise, you have to validate it every time.

devngc:
@mostaphaRoudsari, do I not have to validate every time I make a change to the config.json file?

devngc:
@mostaphaRoudsari, 1. I don't understand how you want me to use Config class to use the config.json file. I thought, we only used Config class to create a validator and also to create schema when we needed it. The following function validates and read a config.json file. I believe we don't have a choice but to validate the config.json everytime a change is made to it.

`def check_config(file_3dm, config_path):
    """Validates the config file and returns it in the form of a dictionary.

    Args:
        file_3dm: A rhino3dm file object.
        config_path: A text string for the path to the config file.

    Returns:
        Config dictionary or None if any of the checks fails.
    """
    # Validate against schema
    Config.parse_file(config_path)

    # Get a config dictionary from the config file
    config = read_json(config_path)

    # Validate config with external binaries
    if check_layers(file_3dm, config) and check_rad(config):
        return config
    else:
        None`

mostaphaRoudsari:
What about this pattern?

```python
def import_3dm(filepath, config_file):

    config = Config.parse_file(config_file)

    # now use this config object!
    load_layers_or_whatever_function(config)

    # and so on!
```

devngc:
Hi @mostaphaRoudsari 
I tried this. The Config object has None values for everything that is not defined and that means I need to modify all the methods that I wrote to work with a config dict. Do you want me to do that?

mostaphaRoudsari:
> I tried this. The Config object has None values for everything that is not defined and that means I need to modify all the methods that I wrote to work with a config dict. Do you want me to do that?

It shouldn't be any different from the JSON object. What if we have a chat about this first thing tomorrow morning?

honeybee_3dm/togeometry.py
```diff
@@ -286,12 +289,25 @@ def solid_to_face3d(brep):
...
         A list of Ladybug Face3D objects.
     """
-    faces = []
+    face3ds = []
+
     for i in range(len(brep.Faces)):
         mesh = brep.Faces[i].GetMesh(rhino3dm.MeshType.Any)
-        faces.extend(mesh_to_face3d(mesh))
+        faces = mesh_to_face3d(mesh)
+        normals = set([face.normal.z for face in faces])
```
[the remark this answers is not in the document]
devngc: I believe this comment is for lines 304 and 306. I introduced tolerance there.

honeybee_3dm/config.py
```diff
@@ -0,0 +1,432 @@
...
+    """Config for layer keys."""
+
+    exclude_from_rad: bool = Field(
+        False,
+        description='Boolean to indicate if this layer should be excluded from '
+        'radiance model.'
+    )
+
+    include_child_layers: bool = Field(
+        False,
```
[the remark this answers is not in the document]
devngc: I have made this key optional itself. 

I would request you to re-think this. Imagine a layer structure like the following image;
layer structure
If we keep to true by default then a lot of child layers will be unnecessary will be imported. If the user does not want it, they will have to change the setting and will have to re-run the script. I imagine the config file to work like a chisel rather than a hammer. Working on specific parts of the file and not touching the rest.

honeybee_3dm/togeometry.py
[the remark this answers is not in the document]
mostaphaRoudsari: It is actually for this line. Right now is the normal vectors are even slightly different the face will be considered nonplanar.

I should also add that this is a wish for the future. It's ok not to do it for now.

honeybee_3dm/config.py
```diff
@@ -0,0 +1,432 @@
...
+
+class Config(BaseModel):
+    """Config for the config file."""
+
+    sources: Dict[str, str] = Field(
+        None,
+        description='Path to the radiance .mat file.'
+    )
+
+    layers: Dict[str, LayerConfig] = Field(
```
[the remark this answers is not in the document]
devngc: Not mentioning a default value makes it a required field. https://pydantic-docs.helpmanual.io/usage/models/#required-fields

honeybee_3dm/config.py
```diff
@@ -0,0 +1,432 @@
...
+    @validator('layers')
+    def check_layers(cls, v):
+        layers = list(v.keys())
+        if not layers:
+            raise ValueError('The config file cannot be used without layer names.')
+        return v
+
+
+def read_json(config_path):
+    """Get a dictionary from a config.json file.
```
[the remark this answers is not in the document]
devngc: You are right. This function is removed since we're using the Config object now.

honeybee_3dm/config.py
devngc: Each layer in the config file is checked against the list of all the layers in rhino. This list includes child layers as well.

honeybee_3dm/config.py
devngc: The way I see it, the validators work on fields. This function checks both the radiance path in `sources` and layer keys in `layers`. It therefore works with two different fields on two different objects (Config and LayerConfig). Please also note that a value (`modifiers_dict`) is passed between two checks in this function.
It seemed complicated to break this function into two validators for two different objects. So I decided to keep as is. Please share if this is unacceptable and I will work on this.

honeybee_3dm/config.py
devngc: I created a value error here but I don't think the value error will ever be raised.

honeybee_3dm/config.py
devngc: I am moving this to another module since this also a method that is using config and other arguments similar to the rest of the helper functions.

honeybee_3dm/config.py
devngc: Moved

honeybee_3dm/togeometry.py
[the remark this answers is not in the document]
devngc: Created and [issue](https://github.com/ladybug-tools/honeybee-3dm/issues/36) to implement later.

honeybee_3dm/config.py
mostaphaRoudsari: > The way I see it, the validators work on fields.

That's not really correct. We use cross reference checks all over the place and it is one of the strengths of Pydantic. See how the `values` are passed after `v` in the first example for validators: https://pydantic-docs.helpmanual.io/usage/validators/ It is called `passwords_match`.

There is also a `root_validator` object: https://pydantic-docs.helpmanual.io/usage/validators/#root-validators but this case is simple enough to just use the `@validator` with `values` argument.

honeybee_3dm/config.py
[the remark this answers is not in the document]
mostaphaRoudsari: Hmmm... I'm not sure why you think it's unnecessary to import them. As a matter of fact they `must` be imported in this specific case. I still think we should import all the child layers by default.

honeybee_3dm/config.py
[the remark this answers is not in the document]
mostaphaRoudsari: I can't see an example which is using `Field` and keeps the default empty as you mentioned. Even if that is the case, I think we should use the mode explicit approach. "Explicit is better than implicit." :)

honeybee_3dm/config.py
```diff
@@ -177,35 +138,40 @@ def check_rad(config):
...
     """
 
-    # Check-04: Check if radiance modifiers can be imported
+    # Check if radiance modifiers can be imported
+    radiance = False
     for layer in config['layers']:
         if 'radiance_material' in config['layers'][layer]:
+            radiance = True
             break
+    if radiance:
```
mostaphaRoudsari: Python provides a `for/else` options.
```suggestion
    for layer in config['layers']:
        if 'radiance_material' in config['layers'][layer]:
            break
    else:
```

honeybee_3dm/config.py
```diff
@@ -233,200 +199,23 @@ def check_config(file_3dm, config_path):
...
         
-        if 'radiance_material' in config['layers'][layer_name]:
-            radiance_modifiers = mat_to_dict(config['sources']['radiance_material'])
-            hb_aperture.properties.radiance.modifier = radiance_modifiers[config[
-                'layers'][layer_name]['radiance_material']]
-            hb_apertures.append(hb_aperture)
+        # Validate config with external binaries
+        if check_layers(file_3dm, config) and check_rad(config):
+            return config
         else:
```
mostaphaRoudsari, line 217: Doesn't the two checks raise and error if something is wrong? Do you need the else clause here?

honeybee_3dm/helper.py
```diff
@@ -26,3 +35,187 @@ def get_unit_system(file_3dm):
...
+        A tuple of grid controls (grid_size, grid_offset) if valid 
+            grid settings are found in the config file or None
+    """
+
+    if 'grid_settings' in config['layers'][layer_name] and \
+            config['layers'][layer_name]['exclude_from_rad']:
+        grid_controls = config['layers'][layer_name]['grid_settings']
+        return grid_controls['grid_size'], grid_controls['grid_offset']
+    else:
+        return None
```
mostaphaRoudsari, line 65: You can remove these lines. Python by default returns None if there is no return.

honeybee_3dm/helper.py
```diff
@@ -26,3 +35,187 @@ def get_unit_system(file_3dm):
...
+        name: A text string of the name of the rhino object.
+        layer_name: A text string of the rhino layer name.
+
+    Returns:
+        A tuple of lists. Each list contains Honeybee Aperture objects, 
+            Honeybee Shade objects, and Honeybee Door objects. List will
+            be empty if no objects are found for that Honeybee object.
+    """
+
+    hb_shades, hb_apertures, hb_doors = tuple([[] for _ in range(3)])
```
mostaphaRoudsari, line 180: What about?

```suggestion
    hb_shades, hb_apertures, hb_doors = ([], [], [])
```

If you still want to use a loop you should remove the extra []:

```python
hb_shades, hb_apertures, hb_doors = tuple([] for _ in range(3))
```

honeybee_3dm/helper.py
```diff
@@ -26,3 +35,187 @@ def get_unit_system(file_3dm):
...
+        hb_door = Door(*args)
+        hb_door.display_name = args[0]
+
+        if 'radiance_material' in config['layers'][layer_name]:
+            radiance_modifiers = mat_to_dict(config['sources']['radiance_material'])
+            hb_door.properties.radiance.modifier = radiance_modifiers[config[
+                'layers'][layer_name]['radiance_material']]
+            hb_doors.append(hb_door)
+        else:
+            hb_doors.append(hb_door)
```
mostaphaRoudsari, line 207: Can't you wrap this code in a function and path the layer name and object to it? The same code is used several times.

honeybee_3dm/helper.py
```diff
@@ -26,3 +35,187 @@ def get_unit_system(file_3dm):
...
+
+    elif config['layers'][layer_name]['honeybee_face_object'] == 'shade':
+        hb_shade = Shade(*args)
+        hb_shade.display_name = args[0]
+
+        if 'radiance_material' in config['layers'][layer_name]:
+            radiance_modifiers = mat_to_dict(config['sources']['radiance_material'])
+            hb_door.properties.radiance.modifier = radiance_modifiers[config[
```
mostaphaRoudsari, line 215: Reusable function will avoid mistakes like this one.

```suggestion
            hb_shade.properties.radiance.modifier = radiance_modifiers[config[
```

mostaphaRoudsari (commented):
Looks good @devngc! See my comments.

honeybee_3dm/config.py
devngc: Complied

honeybee_3dm/config.py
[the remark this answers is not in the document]
devngc: Complied

honeybee_3dm/config.py
devngc: for/else construct will be useful to do something once the loop has finished. Here, we don't want the loop to finish.

mostaphaRoudsari: Let me make sure that I understand the code here. Is it supposed to be for each layer or all the layers? Because the code that you have right now breaks the loop as soon a sit finds a radiance_material key in one of the layers similar to what I have suggested. Or am I missing something here?

mostaphaRoudsari (approved):
Overall it looks good to me. 

I have two comments about `material.py`:

1. In the new libraries, we use `modifier` for radiance `modifiers` and not `material`.

2. I think we have methods in honeybee-radiance that supports from/to file/dict which you should be using instead of writing your own module. @chriswmackey, can you confirm?

Feel free to merge it in as is, and we can refactor the code once @chriswmackey provides more information about working with radiance modifiers.

honeybee_3dm/config.py
```diff
@@ -0,0 +1,196 @@
...
+                raise ValueError(
+                        f'invalid sources key: {sources[0]}.'
+                        'key must be radiance_material.'
+                )
+        return v
+
+    @validator('layers')
+    def check_rad(cls, v, values):
+        config_layers = v
+        sources = values['sources']
```
mostaphaRoudsari, line 109: You have to do a quick check here to ensure the 'sources' key is not None before trying to get the value.

devngc: The value will be none if the path is not provided and I am dealing with the None value.

honeybee_3dm/config.py
devngc: What you understood is correct and that's what I am trying to do. This means I am missing something from what you recommended. I need to look again at for/else.

devngc:
```
animals = ['cat', 'dog', 'human', 'horse']
for animal in animals:
    print(animal)
    if animal == 'human':
        break
else:
    print(animals)
```
In the code above, `animals` will not be printed. which is what we want.

PR #41, opened 2020-12-13
Improve geometry translators and solids to grids

devngc:


honeybee_3dm/face.py
```diff
@@ -38,7 +38,7 @@ def import_objects_with_config(rhino3dm_file, layer, *, tolerance=None,
 
     """
     # Placeholders
-    hb_faces, hb_shades, hb_apertures, hb_doors, hb_grids = tuple([[] for _ in range(5)])
+    hb_faces, hb_shades, hb_apertures, hb_doors, hb_grids = ([], [], [], [], [])
```
mostaphaRoudsari, line 41: Whenever I write code like this, it feels I should be using a dictionary.

honeybee_3dm/togeometry.py
```diff
@@ -166,7 +162,10 @@ def brep_to_meshed_face3d(brep, tolerance):
     for i in range(len(brep.Faces)):
         mesh = brep.Faces[i].GetMesh(rhino3dm.MeshType.Any)
         faces.extend(mesh_to_face3d(mesh))
-    polyface = Polyface3D.from_faces(faces, tolerance)
+    try:
+        polyface = Polyface3D.from_faces(faces, tolerance)
+    except Exception:
+        raise ValueError
```
mostaphaRoudsari, line 168: You should include a helpful error message here. This will be confusing.

raise ValueError('your message here!')

Also what is the type of exception that you're trying to address here? A generic Exception is almost always a bad idea.

PR #48, opened 2020-12-17
Refactor and solids to grids

### LINKED ISSUE #34: Solid Extrusion Brepface Bug
Rhino3dm does not return correct number of brepfaces for a brep created from a solid extrusion. This issue is reported [here](https://mcneel.myjetbrains.com/youtrack/issue/RH3DM-128)

### LINKED ISSUE #35: Explode Brep
Feature request is made [here](https://mcneel.myjetbrains.com/youtrack/issue/RH3DM-130) to facilitate equivalent of brep explode in rhino3dm.

devngc:
This PR contains refactored modules to work with rhino3dm version 0.14.0
resolves #34 
resolves #35

### DIFF
--- a/honeybee_3dm/face.py
+++ b/honeybee_3dm/face.py
@@ -63,8 +63,15 @@ def import_objects_with_config(rhino3dm_file, layer, tolerance, *,
             except AttributeError:
                 raise AttributeError(
                     'Please turn on the shaded mode in rhino, save the file,'
-                    ' and try again.'
+                    ' and try again. Shaded mesh could not be created for'
+                    f' object with ID {obj.Attributes.Id}'
                 )
+            except ValueError:
+                raise ValueError(
+                f'Could not create a face for object of ID {obj.Attributes.Id}'
+                ' Please reduce the unit tolerance value in rhino save the file and'
+                ' try again.'
+            )
             name = obj.Attributes.Name
 
             for face_obj in lb_faces:
@@ -114,6 +121,12 @@ def import_objects(file_3dm, layer, *, tolerance):
                 'Please turn on the shaded mode in rhino, save the file,'
                 ' and try again. Shaded mesh could not be created for'
                 f' object with ID {obj.Attributes.Id}'
+            )
+        except ValueError:
+                raise ValueError(
+                f'Could not create a face for object of ID {obj.Attributes.Id}'
+                ' Please reduce the unit tolerance value in rhino save the file and'
+                ' try again.'
             )
 
         name = obj.Attributes.Name
--- a/honeybee_3dm/face.py
+++ b/honeybee_3dm/face.py
@@ -23,6 +23,7 @@ def import_objects_with_config(rhino3dm_file, layer, tolerance, *,
 
     Args:
         rhino3dm_file: A Rhino3DM file object.
+        layer: A rhino3dm layer object.
         tolerance: A number for model tolerance. By default the tolerance is set to
             the ModelAbsoluteTolerance value in input 3DM file.
         config: A dictionary of config settings. Defaults to None
@@ -45,9 +46,9 @@ def import_objects_with_config(rhino3dm_file, layer, tolerance, *,
     # If Grids are requested for a layer
     if grid_controls(config, layer.Name):
 
-        hb_grids = import_grids(rhino3dm_file, layer,
+        hb_grids = import_grids(rhino3dm_file, layer, tolerance,
             grid_controls=grid_controls(config, layer.Name),
-            child_layer=child_layer_control(config, layer.Name), tolerance=tolerance)
+            child_layer=child_layer_control(config, layer.Name))
 
     # If Grids are not requested for a layer
     else:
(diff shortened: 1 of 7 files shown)


honeybee_3dm/face.py
```diff
@@ -85,13 +103,13 @@ def import_objects_with_config(rhino3dm_file, layer, *, tolerance=None,
...
-                    hb_shades.extend(hb_objects[1])
-                    hb_doors.extend(hb_objects[2])
+                    hb_doors.extend(hb_objects[1])
+                    hb_shades.extend(hb_objects[2])
 
     return hb_faces, hb_shades, hb_apertures, hb_doors, hb_grids
 
 
-def import_objects(file_3dm, layer, *, tolerance=None):
+def import_objects(file_3dm, layer, *, tolerance):
```
mostaphaRoudsari, line 112: I'm curios to know why you are removing the option to use ModelAbsoluteTolerance as the default value. In any case, you should update the docstring.

honeybee_3dm/grid.py
```diff
@@ -44,32 +44,35 @@ def import_grids(rhino3dm_file, layer, *, grid_controls=None, child_layer=False,
...
-                    'A Non-planar Brep is not handled by Honeybee for grids.' 
-                    ' It is ignored.'
-                )
-                continue
-        # If it's a Mesh
-        elif isinstance(geo, rhino3dm.Mesh):
+        
+        # If it's a Mesh use it to create grids
+        # This is done so that if a user has created mesh with certain density
+        # the same can be used to create grids
```
mostaphaRoudsari, line 50: 👍

honeybee_3dm/grid.py
```diff
@@ -44,32 +44,35 @@ def import_grids(rhino3dm_file, layer, *, grid_controls=None, child_layer=False,
...
+            hb_grids.append(SensorGrid.from_mesh3d(*args))
 
-        name = obj.Attributes.Name
-        obj_name = name or clean_and_id_string('Grid')
-        args = [clean_string(obj_name), mesh3d]
-        hb_grids.append(SensorGrid.from_mesh3d(*args))
+        else:
+            for face in to_face3d(obj, tolerance):
+                if face.normal.z == -1:
+                    continue
```
mostaphaRoudsari, line 61: Not sure why you ignore the faces that are looking downwards.

honeybee_3dm/togeometry.py
```diff
@@ -191,151 +185,155 @@ def brep_to_mesh_to_face3d(brep):
...
-        raise AttributeError
-    # * Check 01 - If any of the edge is an arc
-    lines = []
+
+    # If any of the edge is an arc
     curved = False
     for i in range(len(brep.Edges)):
         if not brep.Edges[i].IsLinear(tolerance):
             curved = True
             break
```
mostaphaRoudsari, line 205: I comment on similar cases before. This is a typical use case for using **for**/else instead of creating a Boolean.

honeybee_3dm/grid.py
devngc: Because you said so 😀. You said we create grids from all faces of a brep except bottom.

honeybee_3dm/togeometry.py
devngc:
```
animals = ['cat', 'dog', 'human', 'horse']
for animal in animals:
    print(animal)
    if animal == 'human':
        break
else:
    print(animals)
```
In the code above, `animals` will not be printed. which is what we want. My understanding is that the else clause will only be executed if the loop is finished normally. Whereas here, we want the loop to end abruptly when we find that one of the edge is curved.

honeybee_3dm/grid.py
mostaphaRoudsari: I see! That was a project specific comment not a generic comment to be hard-coded into honeybee-3dm. Let's remove it from here and think of a better way to achieve our goal for the specific project.

honeybee_3dm/togeometry.py
mostaphaRoudsari: I see your point now. You are right that else won't work. In that case you really want to `return` when it happens and separate the rest of your logic after that.

```python

    for i in range(len(brep.Edges)):
        if not brep.Edges[i].IsLinear(tolerance):
            return [brep_to_meshed_face3d(brep, tolerance)]

   # the rest of the code here if the case didn't happen.

```

devngc: Complied

honeybee_3dm/grid.py
devngc: Resolved

honeybee_3dm/face.py
devngc: In the `togeometry` module, there are several methods that use the tolerance value and I observed that when I forgot to provide the tolerance value and used a default value then it created weird results. So I went ahead and made it mandatory to define them.

mostaphaRoudsari (approved):
Thanks @devngc! LGTM.

PR #83, opened 2021-01-13
Export base_geometry  for Sensorgrid objects

devngc:
This PR makes use to recently added base_geometry property and to_face3d method in the Sensorgrid module.

### DIFF
--- a/honeybee_3dm/grid.py
+++ b/honeybee_3dm/grid.py
@@ -56,20 +56,20 @@ def import_grids(rhino3dm_file, layer, tolerance, *, grid_controls=None,
             hb_grids.append(SensorGrid.from_mesh3d(*args))
 
         else:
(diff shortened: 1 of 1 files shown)


mostaphaRoudsari (approved):
Thanks @devngc, I don't understand the details here but overall it looks good. Maybe we should add a test too.

devngc:
This module needs a lot of tests. I thought writing tests is a good candidate for rest cycles that you mentioned on the call.

devngc:
HI @mostaphaRoudsari, Earlier we were meshing all the Face3Ds before adding them as Sensorgrids. Now Chris has added a method to create Sensogrids directly from Face3Ds. So I changed that. This also means that the Sensorgrids generated from Face3Ds will have `base_geometry` as a property. Which we are interested in honeybee-vtk.

PR #85, opened 2021-01-13
refactor(face): Warn instead of stopping the script for smaller geometries

devngc:
Hi @mostaphaRoudsari,
This is as per "Faces with edges MUCH smaller than unit tolerance" in honeybee-3dm on basecamp.

### DIFF
--- a/honeybee_3dm/face.py
+++ b/honeybee_3dm/face.py
@@ -13,7 +13,7 @@ from .helper import grid_controls, face3d_to_hb_face_with_face_type, face3d_to_h
 from .helper import face3d_to_hb_face_with_rad, child_layer_control
 
 
-tolerance_error = 'Could not create a face for object of ID {} Please reduce the unit' \
+tolerance_warning = 'Could not create a face for object of ID {} Please reduce the unit' \
 ' tolerance value in rhino, save the file and try again. You might need to repeat' \
 ' this more than once if the face is too small for the unit tolerance selected.'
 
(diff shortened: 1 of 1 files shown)


honeybee_3dm/face.py
```diff
@@ -70,12 +70,14 @@ def import_objects_with_config(rhino3dm_file, layer, tolerance, *,
...
                     )
             except AssertionError:
-                raise AssertionError(
+                warnings.warn(
                     f'Could not create a face for object of ID {obj.Attributes.Id}'
                     ' Please reduce the unit tolerance value in rhino, save the file and'
                     ' try again. You might need to repeat this more than once if the'
                     ' face is too small for the unit tolerance selected.'
                     )
+                continue
```
chriswmackey, line 79: You can just define this long message once with blank `{}` and use `.format({obj.Attributes.Id})` to fill in the object ID in the different places.

mostaphaRoudsari (approved):
Looks good to me!

honeybee_3dm/face.py
devngc: @chriswmackey, I am not sure if I understand your suggestion. Can you share an example?

mostaphaRoudsari: @chriswmackey is suggesting reusing the text as a template.

```python
warn_msg = 'Could not create a face for object of ID {object_id}' \
                    ' Please reduce the unit tolerance value in rhino, save the file and' \
                    ' try again. You might need to repeat this more than once if the' \
                    ' face is too small for the unit tolerance selected.'
```

Then when you want create a warning you can do.

```python
warning.warn(warn_msg .format(object_id=obj.Attributes.Id)
```

This will avoid writing the same long message twice. I used a variable name (object_id) to make it more readable but you can also use a blank `{}` as suggested by Chris. In that case it will look like this.


```python
warning.warn(warn_msg .format(obj.Attributes.Id)
```

devngc:
Thank you @mostaphaRoudsari and @chriswmackey. I revised as per suggestion.

PR #109, opened 2021-01-27
Add cli

devngc:


mostaphaRoudsari:
Hi @devngc, I suggest to call this command also as `translate` to keep it consistent with the other libraries. In fact, you are translating a Rhino file to a HBJSON file.

